PR #1, opened 2011-04-04
Renamed the readme file to shown automatically on github

lsolesen:


michelf:
Sorry, not acceptable.

For one thing, this needs to be a single commit; it being two commits (one adding a file, one removing a file) makes me worry that git won't see it as a rename and that the file will get disconnected from the readme in the extra branch. Right now, when I change something in the master branch I can go to the extra branch and merge it with master and the version history will be merged, if git doesn't see this as a renamed file it'll break that connection.

Secondly, the file has been named like this for a long time (before github or even git was created), so I'm not too inclined changing it.

PR #4, opened 2011-04-04
Renaming - in one commit!

lsolesen:
Renamed to make it display automatically on github and follow convention from a lot of other projects

I made it into one commit.

I still think the change would be good now you are hosting on github. Whether the file has had the other name for a long time, does not seem to be a good reason not to change it?


michelf:
It might be a good idea for github hosting, but PHP Markdown's primary distribution method is through a zip archive, not through github. If I could tell github which file to display I would, but I don't feel it is worth renaming the file.

Perhaps if one day everything gets repackaged differently it would make sense to do that, but right now I'd rather not.

PR #6, opened 2012-01-12
Add CSS classes to fenced code blocks

coreyworrell:


michelf:
Thanks. This is something I planned on implementing for a while.

Two small issues:
1. This code will write an empty class="" attribute when no class is present. I know it's not a big issue, but this makes the resulting HTML a little dirty.
2. I'd prefer if the syntax used a dot in front of the class name in the Markdown source, as it mimics how classes are represented in CSS.
3. I think it'd be a good idea to let the writer put the class name on the closing fence if he so prefers... something like this:
   
   ```
   ~~~
   some code
   ~~~ .php
   ```

Can you manage to do this?

coreyworrell:
See new pull request. Sorry for the weird branches and multiple requests, I'm not great with Github.

PR #16, opened 2012-02-08
Allow multiple references to the same footnote

smokris:
Currently the following markdown doesn't parse as I'd expect:

```
    - refer to footnote A[^1]
    - refer to footnote B[^2]
    - refer to footnote A again[^1]

[^1]: footnote A
[^2]: footnote B
```

The "refer to footnote A again[^1]" line doesn't get linked to footnote A.

I've modified the footnote callback so that this is possible.

### DIFF
--- a/markdown.php
+++ b/markdown.php
@@ -2748,14 +2748,12 @@ class MarkdownExtra_Parser extends Markdown_Parser {
 	function _appendFootnotes_callback($matches) {
 		$node_id = $this->fn_id_prefix . $matches[1];
 		
-		# Create footnote marker only if it has a corresponding footnote *and*
-		# the footnote hasn't been used by another marker.
+		# Create footnote marker only if it has a corresponding footnote
 		if (isset($this->footnotes[$node_id])) {
-			# Transfert footnote content to the ordered list.
+			# Transfer footnote content to the ordered list.
 			$this->footnotes_ordered[$node_id] = $this->footnotes[$node_id];
-			unset($this->footnotes[$node_id]);
 			
-			$num = $this->footnote_counter++;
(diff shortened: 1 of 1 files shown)


michelf:
This shouldn't work like this. There is no requirement that the footnote reference markers in the Markdown source start from one, be in the right order, or be numbers at all. So you should not rely on the footnote reference marker in `$matches[1]` to be the right number. That's why there was a counter there before.

Otherwise, it's a nice patch. Can you fix that?

michelf:
I implemented it slightly differently so that footnotes markers are properly numbered even if the user didn't use numbers as footnote reference in his Markdown file.

taufik-nurrohman:
When writing multiple footnotes for the same reference, I got this weird reference ID pattern. Is this valid?

#### Example

``` .markdown
Lorem ipsum [^1] dolor sit amet [^2]

Lorem ipsum [^1] dolor sit amet [^2]

[^1]: Footnote 1
[^2]: Footnote 2
```

#### Footnote Output
1. Footnote 1 [↩]() [↩]()
2. Footnote 2 [↩]() [↩]()

#### Hash Produced

``` .no-highlight
#fnref:1
#fnref2:1
#fnref:2
#fnref2:2
```

Why not produce this instead?

``` .no-highlight
#fnref:1
#fnref:1-2
#fnref:2
#fnref:2-2
```

or &hellip;

``` .no-highlight
#fnref:1
#fnref-2:1
#fnref:2
#fnref-2:2
```

michelf:
@tovic This has little to do with this four year old pull request. Please start a new issue if you want to discuss this.

PR #25, opened 2012-04-14
Adding html5 block level elements, see http://drupal.org/node/1366260

barraponto:


michelf:
I definitely want to merge this, I just want to take the time to check it against the full list of HTML5 elements. It might be a good thing for instance to add `<audio>` and `<video>` to list "a" (those items which can be both block or inline), perhaps others fall in that category.

Thank you for your patience.

michelf:
I made the changes slightly differently, but latest HEAD in all branches should support HTML 5 elements.

PR #26, opened 2012-04-20
Merged TOC feature from jtopjian's fork

mytharcher:
I think auto-generation of table of content(TOC) is very useful. And I searched and found jtopjian's fork had done that yet. so I merge his branch to my fork.

Beside jtopjian's TOC upgrade, I also did an improvement, which one could use TOC by headers' ID not specified(the not specified ID would be replaced by the header text).

### DIFF
--- a/License.text
+++ b/License.text
@@ -1,0 +1,36 @@
+PHP Markdown & Extra  
+Copyright (c) 2004-2009 Michel Fortin  
+<http://michelf.com/>  
+All rights reserved.
+
+Based on Markdown  
+Copyright (c) 2003-2006 John Gruber   
+<http://daringfireball.net/>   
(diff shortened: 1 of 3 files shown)


michelf:
I don't think it is the role of Markdown (or Markdown Extra) to create an outline of your document. If you need an outline, it can and should be done by scanning the resulting HTML following the algorithm specified in the HTML5 spec.

Also, auto-generating id attributes is problematic, because they must be unique. The user might have specified the same value for an id attribute elsewhere in the document, or might embed the document into something which already has an identical id attribute.

Otherwise, this patch looks well done. If it works well for you that's great. But I won't merge it for the reasons stated above, sorry.

PR #29, opened 2012-05-03
Modifications to adapt Markdown-Parser towards a post-2005 object orientation

cfv1984:
Mostly consisting of setting class constants and access modifiers. The code base itself could use some brushing up probably, but I'm not that intimate with the parser yet.

### DIFF
--- a/markdown.php
+++ b/markdown.php
@@ -2,6 +2,8 @@
 #
 # Markdown  -  A text-to-HTML conversion tool for web writers
 #
+# Modifications towards php5 OO by cfv1984
+#
 # PHP Markdown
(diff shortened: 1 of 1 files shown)


baldurrensch:
I would definitely also want the Markdown parser to be more encapsulated. One way to achieve this and still keep the backwards compatibility is to move the parser class into its own file (bonus points for namespacing it), leave the defines in place and just use those to set those parameters in the parser class. That should make everyone happy.

michelf:
I think the newly added [lib branch](https://github.com/michelf/php-markdown/tree/lib) is going towards what you want. Right now, it's pre-release. Eventually, this'll probably become the default branch.

PR #30, opened 2012-05-11
Added ability for fenced code block to allow a language to be defined.

alexrussell:
After the "~~~" of the start of the block, the language can come straight away or after some whitespace and then can have any whitespace after it before the newline that ends the fence marker line.

The language itself can be alphanumeric with '_' and '-', and the generated code is either the normal code for the code block pre>code>{content} if no language is present, or pre.prettyprint>code.lang-{lang}>{content} if a
language is specified. This fits in nicely with Google Prettify but also nice HTML5 standards, too.

Feel free to discard the 'coding standards' commit. My text editor trims whitespace so I figured I'd do it as a separate commit in case you wanted to ignore it.

### DIFF
--- a/markdown.php
+++ b/markdown.php
@@ -3,11 +3,11 @@
 # Markdown Extra  -  A text-to-HTML conversion tool for web writers
 #
 # PHP Markdown & Extra
-# Copyright (c) 2004-2012 Michel Fortin  
+# Copyright (c) 2004-2012 Michel Fortin
 # <http://michelf.com/projects/php-markdown/>
 #
 # Original Markdown
-# Copyright (c) 2004-2006 John Gruber  
+# Copyright (c) 2004-2006 John Gruber
 # <http://daringfireball.net/projects/markdown/>
 #
 
@@ -79,7 +79,7 @@ Author URI: http://michelf.com/
 if (isset($wp_version)) {
 	# More details about how it works here:
 	# <http://michelf.com/weblog/2005/wordpress-text-flow-vs-markdown/>
-	
+
 	# Post content and excerpts
 	# - Remove WordPress paragraph generator.
 	# - Run Markdown on excerpt, then remove all tags.
@@ -94,13 +94,13 @@ if (isset($wp_version)) {
 		add_filter('get_the_excerpt', 'trim', 7);
 		add_filter('the_excerpt',     'mdwp_add_p');
 		add_filter('the_excerpt_rss', 'mdwp_strip_p');
-		
+
 		remove_filter('content_save_pre',  'balanceTags', 50);
 		remove_filter('excerpt_save_pre',  'balanceTags', 50);
 		add_filter('the_content',  	  'balanceTags', 50);
 		add_filter('get_the_excerpt', 'balanceTags', 9);
 	}
-	
+
 	# Add a footnote id prefix to posts when inside a loop.
 	function mdwp_MarkdownPost($text) {
 		static $parser;
@@ -115,7 +115,7 @@ if (isset($wp_version)) {
 		}
 		return $parser->transform($text);
 	}
-	
+
 	# Comments
 	# - Remove WordPress paragraph generator.
 	# - Remove WordPress auto-link generator.
@@ -130,7 +130,7 @@ if (isset($wp_version)) {
 		add_filter('get_comment_text',    'Markdown', 6);
 		add_filter('get_comment_excerpt', 'Markdown', 6);
 		add_filter('get_comment_excerpt', 'mdwp_strip_p', 7);
-	
+
 		global $mdwp_hidden_tags, $mdwp_placeholders;
 		$mdwp_hidden_tags = explode(' ',
 			'<p> </p> <pre> </pre> <ol> </ol> <ul> </ul> <li> </li>');
@@ -138,7 +138,7 @@ if (isset($wp_version)) {
 			'pEj07ZbbBZ U1kqgh4w4p pre2zmeN6K QTi31t9pre ol0MP1jzJR '.
 			'ML5IjmbRol ulANi1NsGY J7zRLJqPul liA8ctl16T K9nhooUHli'));
 	}
-	
+
 	function mdwp_add_p($text) {
 		if (!preg_match('{^$|^<(p|ul|ol|dl|pre|blockquote)>}i', $text)) {
 			$text = '<p>'.$text.'</p>';
@@ -146,7 +146,7 @@ if (isset($wp_version)) {
 		}
 		return $text;
 	}
-	
+
 	function mdwp_strip_p($t) { return preg_replace('{</?p>}i', '', $t); }
 
 	function mdwp_hide_tags($text) {
@@ -218,7 +218,7 @@ class Markdown_Parser {
 	# Needed to insert a maximum bracked depth while converting to PHP.
 	var $nested_brackets_depth = 6;
 	var $nested_brackets_re;
-	
+
 	var $nested_url_parenthesis_depth = 4;
 	var $nested_url_parenthesis_re;
 
@@ -229,11 +229,11 @@ class Markdown_Parser {
 	# Change to ">" for HTML output.
 	var $empty_element_suffix = MARKDOWN_EMPTY_ELEMENT_SUFFIX;
 	var $tab_width = MARKDOWN_TAB_WIDTH;
-	
+
 	# Change to `true` to disallow markup or entities.
 	var $no_markup = false;
 	var $no_entities = false;
-	
+
 	# Predefined urls and titles for reference links and images.
 	var $predef_urls = array();
 	var $predef_titles = array();
@@ -245,17 +245,17 @@ class Markdown_Parser {
 	#
 		$this->_initDetab();
 		$this->prepareItalicsAndBold();
-	
-		$this->nested_brackets_re = 
+
+		$this->nested_brackets_re =
 			str_repeat('(?>[^\[\]]+|\[', $this->nested_brackets_depth).
 			str_repeat('\])*', $this->nested_brackets_depth);
-	
-		$this->nested_url_parenthesis_re = 
+
+		$this->nested_url_parenthesis_re =
 			str_repeat('(?>[^()\s]+|\(', $this->nested_url_parenthesis_depth).
 			str_repeat('(?>\)))*', $this->nested_url_parenthesis_depth);
-		
+
 		$this->escape_chars_re = '['.preg_quote($this->escape_chars).']';
-		
+
 		# Sort document, block, and span gamut in ascendent priority order.
 		asort($this->document_gamut);
 		asort($this->block_gamut);
@@ -267,27 +267,27 @@ class Markdown_Parser {
 	var $urls = array();
 	var $titles = array();
 	var $html_hashes = array();
-	
+
 	# Status flag to avoid invalid nesting.
 	var $in_anchor = false;
-	
-	
+
+
 	function setup() {
 	#
-	# Called before the transformation process starts to setup parser 
+	# Called before the transformation process starts to setup parser
 	# states.
 	#
 		# Clear global hashes.
 		$this->urls = $this->predef_urls;
 		$this->titles = $this->predef_titles;
 		$this->html_hashes = array();
-		
+
 		$in_anchor = false;
 	}
-	
+
 	function teardown() {
 	#
-	# Called after the transformation process to clear any variable 
+	# Called after the transformation process to clear any variable
 	# which may be taking up memory unnecessarly.
 	#
 		$this->urls = array();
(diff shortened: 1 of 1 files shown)


michelf:
Take a look at [pull request 7](https://github.com/michelf/php-markdown/pull/7).

alexrussell:
Ahh sorry, I forgot to check existing requests. I personally like my method best so I'm going to stick with that, but good luck on sorting out all issues on the other version of the feature.

I don't think the license specifies that any modifications to the source need to be made public, but I figured it was a good thing to do, so that was my primary reason for forking and uploading. The pull request was a secondary thought in case you fancied extending Markdown Extra with my changes, really, but like I said I didn't pay any attention to existing pull requests.

michelf:
No need to apologize. There is indeed no need for the changes to be public, but I appreciate that you took the time to publish them.

coreyworrell:
If this works without any bugs, maybe you can just adapt it to the style that was proposed in #7.

ghost:
@alexrussell Is your revision of markdown extra working without any issues? I desperately need the ability to define languages on code blocks. And if your answer is yes, [this is the file you use](https://github.com/clevercherry/php-markdown/blob/0f50106a9e2aac6f8defb4b28f3f9fb956d13edd/markdown.php), right?

@michelf and @coreyworrell I understand that you are extremely busy right now, but what would be your suggestion?

michelf:
@Ahan It'll have problems with the HTML block parser which must be able to recognize code blocks to do its work. Which means that HTML-like things in code blocks will do weird things. @coreyworrell's pull request #7 is a little better in this regard, but not quite there yet.

alexrussell:
@Aahan as @michelf said obviously it's not perfect. It did the job for me very well, and I've had no issues, but without any formal testing on the markdown and markdown extra files it's hard to say what's likely to break unless you know it inside and out as @michelf and @coreyworrell obviously do.

Maybe give it a try and see if you can bear with the possible bugs to get the feature you wanted. Or alternatively go with @coreyworrell's method from pull request #7.

But to answer your question I don't have any issues, and yes that's the exact file I use. If line 2541 mentions optional language you're good to go.

alexrussell:
@michelf what is the specific problem here? I see that if you include '<?php' and '?>' (but only when using the ~~~ + language form of fenced blocks) the angle brackets don't get htmlspecialchars'd so I see the error. What it'd be nice to know is where the code that deals with the html blocks and basically overrides this is. It looks like anything in the pre/code tag combo should be being htmlspecialchars'd, and yet these aren't. And I can't find any reference to htmlspecialchars_decode or html_entity_decode in the parser class. I'd like to try and squash this, but can't see where exactly it's being messed with.

I was kinda hoping it'd just be a simple case of changing the HTML-matching regex to incorporate the exact regex change I did for fenced blocks, but it looks like it's not that simple. Any info on the ins and outs of the parser would be useful here.

michelf:
The problem is that PHP Markdown (based on Markdown.pl) is made of many successive regex-based parsers. The HTML block parser runs before everything else, it catches HTML blocks and makes them invisible to the following parsers so that parts of HTML blocks looking like Markdown constructs aren't treated as such by the following parsers.

To do its work correctly, the HTML block parser needs to be aware about some Markdown constructs, especially code blocks, and not escape them. So it needs to be updated for this change.

The HTML block parser finds everything in one pass, and can recurse (allowing you to add Markdown inside HTML blocks). It's much more difficult to work with than the others. It could be improved to parse everything and replace all the other parsers for block-level constructs, which would probably be more efficient and have less edge cases. That'd be an interesting project, but I don't have the time for that right now.

alexrussell:
Well thanks for the info - good to just know that the reason the htmlspecialchars isn't being run on the code inside the fenced block is because as far as the preg callback knows by that time, there is no content (presumably some kind of placeholder ready for re-insertion all un-escaped at the end?). I'll have a look into making the HTML pass not touch code inside fenced blocks regardless of language choice.

alexrussell:
@Aahan I have now [updated my code](https://github.com/clevercherry/php-markdown/blob/715b2f470769604f7d1e6487b9b26959f545f22f/markdown.php) to take account of new-style fenced blocks when looking at HTML blocks. Not sure it does exactly what @michelf was concerned about, but at least it now allows `<?php echo fimage(); ?>`-like code in a fenced-with-language block.

Interestingly I think I may have found a bug in the regular-fenced blocks code though - if there's inline php (like `<?php echo fimage(); ?>`) - this PHP code doesn't get HTML-escaped and ends up going in unaltered (thus looking like an HTML comment/tag to browsers). Interestingly, this behaviour doesn't happen in a fenced-with-language block. Very weird. The sample markdown I'm using is at https://gist.github.com/3025913 - the problem section is labelled "Fenced, inline PHP"

@michelf if you want me to expand the pull request to include the new commits and reopen I'll gladly do so - but I'm sure it's still not 100% perfect. I don't suppose you do have any testing framework in place for this do you?

michelf:
@alexrussell Just uploaded [MDTest](https://github.com/michelf/mdtest).

This is the current working copy of those tests, some of them fail even with the currently released version, because there are some bugs in need for a fix. You'll probably be interested in those relating to fenced code blocks.

alexrussell:
It's definitely not a bug in @coreyworrell's fork, as I forked @michelf's repo directly and made my own edits. However, it looks like maybe it's a general bug with any type of fenced-with-language implementation if we've both created the same bug. I was about to say I'm happy with my current implementation, but actually in a different project but using the exact same markdown parser I get different (and not good) results, so I think I'll have to revisit this.

However, it's really not a priority for me, as I don't care about inline HTML in the markdown code and I know to not put the `<?php` and `?>` in my code when I want PHP highlighted - just put the raw PHP in.

alexrussell:
@michelf ahh thanks for that. I'll give them a run over my fork.

michelf:
@Aahan Assuming it works the way I expect it to work (see discussion in pull #7), that it does not introduce a regression (passing the test suite is a good test for that), and that the new feature is well tested (adding some tests for it to MDTest), I see no reason why I wouldn't accept it. But that's still a non-negligeable amount of work.

I intended to look into this myself sometime this summer, basing myself on @coreyworrell's work (pull #7) and then fixing anything that doesn't work. But if someone can beat me to it, or do some parts in advance, that'd be great.

alexrussell:
@Aahan @michelf let's not get ahead of ourselves. Majorly bombed out on the tests. In fact it apparently fails ALL Markdown Extra tests (although the fenced blocks test failure is because of some XML parse error)! But it does randomly fix the 'emphasis' failure in the 'PHP Markdown' section that Markdown extra fails. No idea how it manages that.

I'll have to look in to why it appears to fail all the Markdown extra tests, as I can't imagine it should fail quite so many quite so badly, but don't have time right now.

alexrussell:
Looking more closely at this, my edits don't appear (i.e. to the eye) to break any of the Markdown Extra tests. Thus I'm looking at either a bug in MDTest (unlikely) or a bug in the way I added my version of Markdown extra into the test framework (who knows?)

It'd be obvious if it failed all tests, just to pass some and fail others something's not quite right, but not easy to work out.

But like I said, the actual HTML rendering looks fine to me (but maybe I'm missing something). Even the 'tricky' parts, so I'm not sure what's going on with the weird results when plumbed into MDTest.

What I'll do first is add a few more tests to the MDTest tests (can I say test a few more times?) in fenced code blocks and have a look at what it says for both original Markdown Extra and the version with my additions. The problem with MDTest is that it performs all tests for a given area in one go, and if any one fails it calls the section failed. It'd probably be better to judge each sub-test on its own merit, although I understand that half the testing can very well be in there for interoperability in the bundle between different areas of the syntax.

alexrussell:
Found my problem: I'm an idiot. I'd cloned my repo and copied 'markdown.php' in as my version without first switching to the 'extra' branch.

So yes, it passes and fails all the tests that normal markdown extra does (yay), so now I'll add these extra tests in for crazy php/html tags inside code blocks, etc.

alexrussell:
As I said, it passed the same tests that regular markdown extra passes. But that's not to say it's perfect - markdown extra fails some of the tests anyway, and mine fails the same ones obviously.

Also I have yet to add the tests that cover the specific problems from PR #7 to see if mine's any good at those issues.

ghost:
> So yes, it passes and fails all the tests that normal markdown extra does (yay), so now I'll add these extra tests in for crazy php/html tags inside code blocks, etc.

Oh, I misinterpreted that somehow. Great to hear. Thanks for your time. :)

alexrussell:
No probs. Hopefully I can get on the additional tests tonight. I know for a fact it'll fail one of them (see [my comment above](https://github.com/michelf/php-markdown/pull/30#issuecomment-6692869)), but it may possibly be a bug in normal markdown extra too, so not the largest deal.

alexrussell:
Right, a bit later than I planned but that's how it goes, eh?

I have added some tests to the mdtest repo, they can be found at [my fork](https://github.com/clevercherry/mdtest) and I've submitted a [pull request](https://github.com/michelf/mdtest/pull/1) to the original mdtest repo too. (I haven't added my version of markdown extra, but that can be found [here](https://github.com/clevercherry/php-markdown/blob/extra/markdown.php)

In terms of results, markdown-extra obviously fails any tests that are related to fences-with-language, but my additions pass all tests on that front.... except:

One of the tests I added tests block-level HTML directly in a fence (language = HTML) and it fails to HTML-encode the text. I'm aware this is one of the primary concerns of @michelf in this PR (and PR #7), however the `<?php echo fimage(); ?>`-type tests pass fine with my language-specific edits. I'll have a look at this HTML example, but if someone could download my version of mdtest and my version of markdown-extra and have a look at what's what it'd be appreciated. For more detail the idea is this:

The test just tests block-level HTML in a regular fence:

```
~~~
<div id="some-div">
    <strong>Hello</strong>, <em>world</em>!
</div>
~~~
```

And it works fine. But this test does not:

```
~~~ html
<div id="some-div">
    <strong>Hello</strong>, <em>world</em>!
</div>
~~~
```

As such, it's definitely a bug in my stuff, and like I said I'll check it out.

ghost:
> @alexrussell said: I'll have a look at this HTML example, but if someone could download my version of mdtest and my version of markdown-extra and have a look at what's what it'd be appreciated.

Hey, did you get time to do it?

alexrussell:
Sorry, I haven't been able to do anything since my last comment. As far as I can tell, it looks like it happens only when the fenced block has a language applied, and when the contents is all HTML and valid and whatever.

To be entirely honest, I think the idea of being able to embed HTML in markdown is asking for trouble regardless of this fenced code blocks situation, and I don't personally use this feature so it's not a high priority for me I'm afraid. I'd almost be temped to pull the HTML feature from my own Markdown extra if this starts becoming an issue for me in fenced blocks and not easily resolved.

ghost:
If the problem is only with HTML, I think I can do away with it.

alexrussell:
It certainly appears to be HTML-only. I added some PHP/HTML tests to the mdtest test bundles and the only one that fails is one where I do a fence block with language and the fenced block contains HTML only (and specifically, starts and ends with HTML (i.e. HTML embedded in text is fine)).

Download [my fork of mdtest](https://github.com/clevercherry/mdtest), add in [my Markdown Extra version](https://github.com/clevercherry/php-markdown/blob/extra/markdown.php) and take a look for yourself - passes all the same tests as markdown extra, passes all but one of my new tests (the HTML-only one).

But yeah, I dunno how easy it is to remove the HTML-parsing ability from Markdown Extra, my guess is 'relatively easy', but you never know. It might be very ingrained into the parser.

michelf:
@alexrussell Extra easy: http://michelf.com/weblog/2007/php-markdown-no-markup/

alexrussell:
A-ha. I'll try to enable this mode and see if the tests all pass tonight.

Interesting reading too, I didn't realise it was part of the Markdown spec to allow HTML. I must admit I don't know a whole lot about Markdown save for my use of it on Github and these occasional bits in my own sites.

ghost:
@alexrussell Just came across this one. This:

```
~~~
<ul>
    <li>Fantastic</li>
    <li>Great</li>
</ul>
~~~
```

Shows up like this:

Code block

And as you can see, I haven't defined the language. I am not sure, but from your previous comments, you don't seem to be aware of this?

**EDIT:**

@michelf Uh oh, the issue seems to be with the original PHP Markdown Extra (and probably thus inherited by Alex's version). I don't think this has been discussed before, has it?

**EDIT:** Okay, hmm... looks like the same HTML parser issue. Sorry if I disturbed you. :)

michelf:
I wasn't aware of this, although I'm not surprised some cases are slipping by due to the way things are implemented. I just added an issue for it.

In this case you can add a paragraph or pretty much anything else before the code block and it'll work fine.

PR #36, opened 2012-07-06
Extra: fixed typos in comments. 

JanX2:


dunn:
One more: "is found whitin a tag" (should be "within") on line 1798.  And John Gruber's name is misspelled in the GitHub description.

PR #38, opened 2012-07-31
Support for header classes

mmcclimon:
I added support for header classes in the Extra with the syntax {.class}, in addition to the id attribute {#id}.

(This is my first attempt at submitting anything on GitHub...sorry if I'm doing it wrong!)

### DIFF
--- a/markdown.php
+++ b/markdown.php
@@ -2226,27 +2226,27 @@ class MarkdownExtra_Parser extends Markdown_Parser {
 	# Redefined to add class attribute support
 	#
 		# Setext-style headers:
-		#	  Header 1  {#header1} {.class1}
+		#	  Header 1  {#header1 .class1}
 		#	  ========
 		#  
-		#	  Header 2  {#header2} {.class2}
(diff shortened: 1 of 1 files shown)


michelf:
Your pull request is excellent. But I think if classes get supported, multiple classes should be allowed and they should be all inside the same {curly brackets} together with the id too. So I don't think I'll accept it in the current form.

mmcclimon:
Good call! I changed some things, and now I think it works fully. Now both id and class attributes for headers are done in the same set of curly braces. All of the following syntaxes work, for both header flavors, as far as I can tell:
- `{#id}`
- `{.class1}`
- `{#id .class1}`
- `{#id .class1 .class2}`
- `{#id.class1.class2}` (no spaces)

The id attribute does have to come first, but otherwise it's tolerant of any number of spaces before/after. Thanks!

michelf:
This was merged yesterday, but I redid parts of the attribute parser so that you don't necessarily have to put the id attribute first.

Your patch was a great base for the more generalized extra attribute block now used for fenced code blocks too. Thanks.

Issue #31, opened 2012-05-25
Composer Package


joseym:
Would be awesome if this were available as a composer package.

javiereguiluz:
I'd also like to use `php-markdown` library as a composer package. Currently, the only available package is a stripped port of `php-markdown`: http://packagist.org/packages/dflydev/markdown

Creating a package is as easy as adding the following `composer.json` file in your project and submitting it to http://packagist.org website:

```
{
    "name": "michelf/php-markdown",
    "type": "library",
    "description": "PHP Markdown",
    "homepage": "http://michelf.com/projects/php-markdown",
    "keywords": ["markdown"],
    "license": "BSD-3-Clause",
    "authors": [
        {
            "name": "Michel Fortin",
            "homepage": "http://michelf.com"
        },
        {
            "name": "John Gruber",
            "homepage": "http://daringfireball.net"
        }
    ],
    "require": {
        "php": ">=4.0.5"
    },
    "extra": {
        "branch-alias": {
            "dev-master": "1.0-dev"
        }
    }
}
```

sandeepshetty:
+1 for making this a composer package. In the meantime here is how I use it with composer:

```
{
    "repositories": [
        {
            "type": "package",
            "package": {
                "name": "michelf/php-markdown-extra",
                "version": "1.2.5",
                "dist": {
                    "url": "http://littoral.michelf.ca/code/php-markdown/php-markdown-extra-1.2.5.zip",
                    "type": "zip"
                }
            }
        }
    ],
    "require": {
        "michelf/php-markdown-extra": "1.2.*"
    }
}
```

mauris:
@michelf Michel please take a look this issue. Over at dflydev's "fork" of your markdown classes, @simensen has done a great job in providing Composer integration and PSR standards compliance. 

I am proposing new changes to bring development between dflydev/dflydev-markdown and michelf/php-markdown together and provide ease of integration into end-user projects through Composer. 

I urge @michelf to take a look at the lengthy discussion developers have made about helping your markdown classes to be a great plugin and library for reusability. See https://github.com/dflydev/dflydev-markdown/pull/5

ghost:
It's good except for the strange use of lower-case. But yes, with @michelf
rather delaying that is a simple task, others are just going to pick up the
ball on this.... and have it seems.

mauris:
@drak - take a look at my pull request. I've made changes to ensure full PSR-0 and PSR-1 compliance. 

[...]

PR #42, opened 2012-08-21
Added composer.json file as described in Issue 31

baldurrensch:
https://github.com/michelf/php-markdown/issues/31

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,26 @@
+{
+    "name": "michelf/php-markdown",
+    "type": "library",
+    "description": "PHP Markdown",
+    "homepage": "http://michelf.com/projects/php-markdown",
+    "keywords": ["markdown"],
+    "license": "BSD-3-Clause",
+    "authors": [
+        {
+            "name": "Michel Fortin",
+            "homepage": "http://michelf.com"
+        },
+        {
+            "name": "John Gruber",
+            "homepage": "http://daringfireball.net"
+        }
+    ],
+    "require": {
+        "php": ">=4.0.5"
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-master": "1.0-dev"
+        }
+    }
+}


marijn:
:+1:

michelf:
Looks like something that'd fit better with the new [lib branch](https://github.com/michelf/php-markdown/tree/lib) I added today. That branch should conform to PSR-0, assuming I haven't done anything wrong.

simensen:
This composer.json will not work as it does not specify the autoloader information required to actually use the classes. I believe the following would need to be added in addition to rebasing agains the lib branch in order for the classes to be usable:

```
"autoload": {
    "psr-0": { "michelf": "" }
},
```

@michelf I'm super excited you are moving in this direction. There are a lot of people using my port ([dflydev/markdown](https://github.com/dflydev/dflydev-markdown)) and I'd be really excited to be able to reference the official php-markdown distribution as an option instead of maintaining my port indefinitely.

I'd be happy to help on any and all fronts with this! Please let me know if you have any questions on Composer, Packagist, or want help maintaining this package on Packagist once you're ready to publish it there.

ghost:
@michelf - yes, a PSR-0 version would be better. You can always make it BC by having a `markdown.php` in the root which `require`'s the other two files.

simensen:
@michelf How much work have you been abel to do on the lib branch? Are you close to being able to release that?

michelf:
I haven't worked much on it. I plan to finalize, annouce, and release it at the same time the next version comes in... Which should be done in less than a month. Technically, there isn't much left to do. Perhaps only adding the composer file.

michelf:
@drak Backward compatibility on the PHP API side is of no concern in the lib branch. It's a clean break. That's the whole point.

I still need to be able to merge with the other branches however, until they are deprecated. So this puts limits on code refactorings. But the API is a clean break.

ghost:
You know for the php 4 version you can just add a files: require and be done with it.... then worry about the PSR-0 stuff on your `lib` branch. 

```
"autoload": {
    "files": ["markdown.php"]
}
```

composer.json
```diff
@@ -0,0 +1,26 @@
...
+    "authors": [
+        {
+            "name": "Michel Fortin",
+            "homepage": "http://michelf.com"
+        },
+        {
+            "name": "John Gruber",
+            "homepage": "http://daringfireball.net"
+        }
+    ],
```
ghost, line 17: Missing compatible loader:

```
"autoload": {
    "files": ["markdown.php"]
}
```

jason-gu:
Please add composer support, thanks

michelf:
This has been merged on the lib branch (as discussed earlier). Note, that as this pull request was made against master, github won't close it automatically and won't show it as merged, so I have to close it manually.

PR #61, opened 2013-01-24
Correct composer autoloader definition.

ghost:
When the namespace was changed, the composer definition needed to be updated too.


chdemko:
+1

Issue #77, opened 2013-03-27
Support for Github-flavored code block


wontonst:
It would be nice to add github-flavored code block style, such as

<pre>
```php
my code
```
</pre>

michelf:
I think this is a poor syntax because it's a valid code span identical to `php my code`. You can use the same syntax replacing ``` with `~` to create a fenced code block (and this syntax works on github too).

PR #84, opened 2013-04-09
#77 support for github-flavored code block.

rbewley4:
Reconsider support for github-flavored code blocks. This syntax is very popular and widely used, thanks to github, so I think it is critical that your library supports it.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -2470,6 +2470,12 @@ class _MarkdownExtra_TmpImpl extends \Michelf\Markdown {
 	# ~~~
 	# Code block
 	# ~~~
+	# 
+	# or
+	# 
+	# ```
+	# Code block
+	# ```
 	#
 		$less_than_tab = $this->tab_width;
 		
@@ -2478,6 +2484,8 @@ class _MarkdownExtra_TmpImpl extends \Michelf\Markdown {
 				# 1: Opening marker
 				(
 					~{3,} # Marker: three tilde or more.
+					|
+					`{3,} # Marker: three or more grave accents (github flavored markdown).
 				)
 				[ ]*
 				(?:


michelf:
Actually, this patch won't work correctly if you have HTML-like content inside the code block. The HTML block parser has to be aware of the code block syntax to work correctly, so it needs to be updated to go with this syntax too. You should use the test cases for fenced code blocks found in MDTest. (Yes, I need to refactor all this to avoid this redundant parsing, I'm just not there yet.)

I know it's easy to implement given the existing fenced code block implementation. I know it's used everywhere on Github. How easy something is to do does not necessarily correlate with whether it should be done however. There's already two syntaxes for code blocks, the threshold for adding another should be pretty high. You'd probably be more convincing if you explained why you need it so desperately.

rbewley4:
Sorry, I didn't realize that you had tests for this library (it's not mentioned in the README file). Are you referring to [this](http://github.com/michelf/mdtest) project? I admit that I did not consider that case. I'll see if I can get it to work properly with your tests, and add some new ones if necessary. 

The reason I think this feature needs be supported is because of the popularity and ubiquity of github's markdown syntax.  Honestly, I rarely see a markdown document that uses tildes to identify code blocks. I know that you were the first to introduce a fenced code block syntax, and it sucks that not everyone adopted it, but that's no reason not to support both. Apparently other users of your library also have the same use case.

I realize that this issue is moot if the user is going to be writing documents from scratch (they can just follow your syntax and be done with it); however, this is a big deal when converting content from a third party, who may have used github's syntax.

rbewley4:
Your test package is out-of-date (uses versions from 8 Jan 2012), and has a horrible pass/fail ratio:
- 35 passed
- 23 failed

michelf:
Indeed. I should probably remove the old versions of PHP Markdown from the test suite project, as I haven't kept it in sync.

I can tell you that the current version of PHP Markdown does pass all the tests. The only exception is one test in the PHP Markdown test suite the Extra parser parses differently on purpose (underscore emphasis).

It also is designed to work with the old packaging model, not what is currently on the lib branch. So right now you need to create a php file with a Markdown function that calls the parser's defaultMarkdown static method. I know it's not ideal, this is a transition period and not everything fit perfectly in place yet. Feel free to improve it.

michelf:
Backtick-fenced code blocks are going to be part of the next release. Only, I'm going to go with pull #132 instead, because it also updates the HTML block parser to correctly work around these new code blocks.

michelf:
Merged #132. The feature is in now.

PR #97, opened 2013-05-11
Add interface for Markdown class

itafroma:
This is a small change that adds a MarkdownInterface to codify the public API and aid with decoupling and dependency injection for consuming projects.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -17,7 +17,7 @@ namespace Michelf;
 # Markdown Parser Class
 #
 
-class Markdown {
+class Markdown implements MarkdownInterface {
 
 	### Version ###
 
--- a/Michelf/MarkdownInterface.php
+++ b/Michelf/MarkdownInterface.php
@@ -1,0 +1,37 @@
+<?php
+#
+# Markdown  -  A text-to-HTML conversion tool for web writers
+#
+# PHP Markdown
+# Copyright (c) 2004-2013 Michel Fortin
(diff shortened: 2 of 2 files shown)


michelf:
Hum, this increases the dependency on PSR-0 autoloading (currently if someone doesn't want to bother with autoloading he can just include Markdown.php and have it work). I can see the usefulness though. I think I'll probably merge this.

davidwindell:
I like the idea, but I don't think the static method belongs in the interface.

jjok:
@michelf The README already says, "this library package is meant to be used with class autoloading. For autoloading to work, your project needs have setup a PSR-0-compatible autoloader."

michelf:
@jjok "Meant to be used with" doesn't necessarily mean "must absolutely be used with". If someone just want to use the parser in his project and it works to just include the file, I'm not going to force him to setup autoloading. Anyway, as I said I intend to merge this. There's a Readme.php example file for a reason.

@davidwindell Why don't you like the static method?

stephen-hill:
I really like this pull request. @michelf, any idea when this pull request will be merged?

michelf:
I thought about it some more. The only "problem" I see is that it makes things more complicated to those who don't want to use autoloading. But that's simple to fix: I'll add a "master" file to include that'll include all the others; those who don't want to mess with autoloading can include that file instead.

So I'll merge this as soon as I've some time to put into PHP Markdown, probably next week.

PR #99, opened 2013-05-12
add auto transfer inner link mode

picasso250:
if this option is set to true, the implicit link [text] not after by
[id] or () will be transform to [text](text)

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -59,6 +59,9 @@ class Markdown {
 	public $predef_urls = array();
 	public $predef_titles = array();
 
+	# auto transfer implicit link [text] to inner link
+	public $auto_inner_link = false;
+
 
 	### Parser Implementation ###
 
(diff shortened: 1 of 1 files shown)


michelf:
I'd tend to believe this will create a lot of unintended links for users. It's also not very useful in the common case, and even though it'd be useful on a wiki I don't think this is the right syntax.

picasso250:
Thank you for peplying. 
My intention is to use that mode in a wiki, and I'm using it in my own wiki. I agree with you that it will create lots of unintended links. So, it's not a good syntax sugar. But what if we auto expend **[text]()** to **[text](text)**, and this mode's name change to _auto fullfill empty link_. To get a _same page link_, users can achieve this by **[text]( )**.
Also and again, this mode is off on default, user must turn it on their own will.
I'm not sure this time is it a good idea. Hope for your respond.

Issue #98, opened 2013-05-11
Adopt PHPdoc-style commenting


itafroma:
To help with IDE integration, provide an easy way to generate API documentation, and generally conform better with how many other PHP projects are commented, I'd like to propose that Markdown adopt [PHPDoc](http://en.wikipedia.org/wiki/PHPDoc)-style tags and C-style comments.

Namely, this would entail adopting the following tags:
- [@author](http://www.phpdoc.org/docs/latest/for-users/phpdoc/tags/author.html)
- [@api](http://www.phpdoc.org/docs/latest/for-users/phpdoc/tags/api.html)
- [@copyright](http://www.phpdoc.org/docs/latest/for-users/phpdoc/tags/copyright.html)
- [@license](http://www.phpdoc.org/docs/latest/for-users/phpdoc/tags/license.html)
- [@param](http://www.phpdoc.org/docs/latest/for-users/phpdoc/tags/param.html)
- [@property](http://www.phpdoc.org/docs/latest/for-users/phpdoc/tags/property.html)
- [@return](http://www.phpdoc.org/docs/latest/for-users/phpdoc/tags/return.html)
- [@see](http://www.phpdoc.org/docs/latest/for-users/phpdoc/tags/see.html)
- [@version](http://www.phpdoc.org/docs/latest/for-users/phpdoc/tags/version.html)

and converting the Perl-style `#` comments to `//` for single line commends and 

``` php
/**
 *
 */
```

for comment blocks.

I'm willing to submit a pull request for this, but it's a fair amount of work and I'd like to see if you'd be amenable to what will amount to a pretty drastic change to how the lib is commented.

itafroma:
Whoops, this is a dupe of #2. Sorry about that!

michelf:
Actually I'd rather have this issue open than #2, since this one is more focused only on documentation comments and is more detailed.

As I said in #2, I'm open to it. Here's my suggestion: do it only for documentation comments of public members (methods and variables) and the class itself. The benefit for public members is much greater (protected members are only meant for tinkering with the parser, which not many people do and the API for that is not considered stable). There's also the issue that I'm still keeping in sync the master and extra branches, and changing comments everywhere would make merges difficult (unless we backport the changes to these branches too). I'll be much more amendable to global changes in style once I drop support for master and extra (next year).

itafroma:
Sounds fair. Just so I'm clear, are you suggesting that:
[...]

PR #100, opened 2013-05-14
Convert public member comments to PHPDoc format

itafroma:
This is the promised pull request discussed in #98. This may be largely academic and not worth merging, however. It turns out there is [a PSR draft proposal](https://github.com/phpDocumentor/phpDocumentor2/blob/develop/docs/PSR.md) being worked out by the phpDocumentor team which may be worth adopting instead whenever it becomes a full PSR.

It's been in development for a couple of years now with no ETA on even a proposal, but it is intended to obsolete the PHPDoc "de-facto" standard, there's been some movement on the FIG side to get a documentation standard going, and it seems likely that the FIG will eventually adopt a documentation standard within the coming months.

So, given the timeline of removing legacy bits of Markdown that would prevent a wholesale conversion of all the comments, it may be worth postponing this entirely to see what the FIG  does in the meantime.

All that said, I still created the PR for informational purposes and a jumping off point for discussions about how best to convert the comments whether it's done now or after a PSR emerges.

As suggested, only comments on public members were converted. I interpreted that as follows:
- The file-level comments were converted over. I couldn't find any guidance on how to handle multiple, disparate copyright tags, so the use of parens to denote John Gruber's copyright claim is definitely malleable.
- The class-level comments for `Markdown`, `MarkdownExtra`, and `_MarkdownExtra_TmpImpl` were converted over. The first two are uncontroversial. I converted `_MarkdownExtra_Tmplmpl` because, while an internal class, it contained public members. I marked it with the `@internal` tag to help avoid ambiguity by people relying on IDEs and such to determine what's usable.
- The `MARKDOWNLIB_VERSION` constant was converted over.
- The constructor comments were converted over under the premise that they are technically public method, even if their usage is either unnecessary or self-documenting due to lack of parameters.
- I made the call to convert the comments within public methods over as well. My reasoning for this was to avoid mixing commenting styles within a single logical block, but I'm not wedded to it.
- Where I saw an opportunity, I removed phrases or sentences that duplicated a tag. An example is changing _Change to `true` to disallow markup or entities_ to simply _Disallow markup_ because the `@var boolean` tag indicates the allowed values. 

As an aside, in writing this, it was—for lack of a better word—weird to have two styles of commenting mixed together in the same file. Barring the PSR issue, that alone may be worth postponing this issue until everything can be converted over.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -1,62 +1,132 @@
 <?php
-#
-# Markdown  -  A text-to-HTML conversion tool for web writers
-#
-# PHP Markdown  
-# Copyright (c) 2004-2013 Michel Fortin  
-# <http://michelf.com/projects/php-markdown/>
-#
-# Original Markdown  
-# Copyright (c) 2004-2006 John Gruber  
-# <http://daringfireball.net/projects/markdown/>
-#
+
+/**
+ * Markdown - A text-to-HTML conversion tool for web writers
+ *
+ * @copyright 2004-2013 Michel Fortin <http://michelf.com/projects/php-markdown/>
+ * @copyright (Original Markdown) 2004-2006 John Gruber <http://daringfireball.net/projects/markdown/>
+ *
+ * @license http://michelf.ca/projects/php-markdown/#license
+ */
+
 namespace Michelf;
 
-
-#
-# Markdown Parser Class
-#
-
+/**
+ * Markdown Parser Class
+ */
 class Markdown {
 
-	### Version ###
-
+	/**
+	 * Version
+	 * @var string
+	 */
 	const  MARKDOWNLIB_VERSION  =  "1.3";
 
 	### Simple Function Interface ###
 
+	/**
+	 * Initialize the parser and return the result of its transform method.
+	 *
+	 * This will work fine for derived classes too.
+	 *
+	 * @api
+	 *
+	 * @param string $text The Markdown text to transform.
+	 *
+	 * @return string The transformed string in HTML.
+	 */
 	public static function defaultTransform($text) {
-	#
-	# Initialize the parser and return the result of its transform method.
-	# This will work fine for derived classes too.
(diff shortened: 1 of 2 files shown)


michelf:
Looks good. You could do a little better by transcribing some text from here:
http://michelf.ca/projects/php-markdown/configuration/

As for the "change to true" thing, I prefer redundancy over ambiguity. Acknowledged, there's not much ambiguity given the variable name, but still, setting to true to disable something might be somewhat counterintuitive. Also, with two-word documentation like that you're pretty much just repeating the variable name (replacing "no" with "disallow"), which makes it pointless. At least that's how I see it.

Your choice of what to convert is pretty much what I'd have done myself, except perhaps for comments inside the constructors because they're in common with the other branches. But given this is only two functions and they're unlikely to change I won't mind one or two merge conflicts if it occurs eventually.

As for postponing conversion, perhaps that's a good idea. I know I can cope with two comments styles, but I can see how I might get negative feedback over this.

itafroma:
Thanks for the feedback.

I agree with you on effectively duplicating the variable name in the docblock and I think expanding the docblocks for properties with the web documentation will help to address this.

I've just amended the PR to include the web documentation: the two word "disallow markup" and "disallow entities" lines are still there to provide the required one line short description of the property, but the expanded transcribed documentation afterwards effectively brings back the "change to true" directives.

robbieaverill:
Yes please!!

itafroma:
I encourage anyone else to take up this effort if it's still warranted, but this is way out of date.

michelf:
If someone wants to do the work of updating the comments, feel free to do so in a new pull request. I'll gladly accept it now that I'm no longer keeping the old classic version in sync.

robbieaverill:
Hey guys. I'm happy to do this, have made a bit of a start in my fork's `issue/98-phpdoc-comments` branch and will chip away at it over the next couple of days.

PR #101, opened 2013-05-16
transform file / id's classes for blockquotes

WickedSik:
- Allowing ID's and classes to be added to blockquotes as well.
- added a transformFile method


michelf:
This pull should really be two pulls. Beside that, I'm not sure there's that much of a benefit in having a transform file function. I'm sure many people will interpret that as generating a full HTML file, which it doesn't, and it can't really do it either, not in a way that'll be useful to most people without a template system.

As for blockquotes with class and id, I'm not against the idea, but the solution isn't so clear. It needs to be coherent for all block-level elements (lists, paragraphs, tables...). It needs to make sense when you read it. And I'd need to check what other Markdown implementations do and decide if I should mimick them to avoid fragmentation. Basically I need to do some research before I can decide what should be done with that.

PR #102, opened 2013-05-24
Unit tests+header option

Znarkus:
Hi! I started work on unit tests (phpunit). Also added configuration option to change header elements.


michelf:
The test suite already exists here: https://github.com/michelf/mdtest
Read my comment here if you want to improve it: https://github.com/michelf/mdtest/pull/3

mtibben:
It would be nice to have the tests in this repo, or at least linked in the Readme. My first thoughts when coming across this lib was "Oh god, there is no test suite".

Also, it looks like it would be fairly trivial to use your existing tests with PhpUnit

michelf:
There a link on the Readme under Development and Testing (added after this issue was first opened):
https://github.com/michelf/php-markdown#development-and-testing

The reason it's a separate tool is to allow it to be used to test others implementations which aren't necessarily in PHP. Also a plain string equality test (as performed in this pull) isn't the ideal comparison tool: MDTest performs some normalization to ignore insignificant whitespace and attributes ordered differently.

michaelbutler:
This could either be deleted or moved to the `test/` directory that was recently created.

PR #112, opened 2013-07-16
Implemented some of Enhanced Ordered Lists (issue 88)

mattgorle:
Numerically ordered lists now start with the first number given.

For example, if you start your list in Markdown as follows:
1. second list item
2. third list item

then Markdown will produce this:

<ol start="2">
<li>second list item</li>
<li>third list item</li>
</ol>

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -851,16 +851,27 @@ class Markdown {
 		$marker_ul_re  = '[*+-]';
 		$marker_ol_re  = '\d+[\.]';
 		$marker_any_re = "(?:$marker_ul_re|$marker_ol_re)";
+		$marker_ol_start_re = '[0-9]+';
 		
 		$list = $matches[1];
 		$list_type = preg_match("/$marker_ul_re/", $matches[4]) ? "ul" : "ol";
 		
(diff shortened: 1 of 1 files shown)


michelf:
While I do support the feature, this could break some documents. Not many Markdown implementations support that, as you can [check on Babelmark](http://johnmacfarlane.net/babelmark2/?normalize=1&text=2.+test%0A3.+test). I'd tend to add a configuration variable and make it enabled only for MarkdownExtra. (The configuration variable would be available on the vanilla Markdown parser but the default value would be `false` in this case.)

Also, your implementation appends `start="1"` in some cases it shouldn't. It should'd add this attribute for unordered lists (I haven't tested but it seems like it'd). Neither should it add the attribute if the value is `"1"` on ordered lists: for one thing it's not needed and also it would force unit tests in MDTest to be rewritten to account for that.

mattgorle:
There are now checks to not add <code>start="1"</code> to anything.  <code>ul</code> is completely ignored and has nothing added to it.

The new behaviour is disabled by default when using the Markdown parser, but can be enabled by setting <code>enhanced_ordered_list</code> to <code>true</code>.  

MarkdownExtra enables the new behaviour by default, by setting <code>enhanced_ordered_list</code> to <code>true</code> in its constructor.

PR #114, opened 2013-07-23
Added transform tests.

olvlvl:
Some transform tests.


olvlvl:
I just realized that you provide unit tests in another repository: https://github.com/michelf/mdtest

This is really confusing because unit tests are usually available in the main repository. This is especially confusing because there is no mention of that repository in your README. You really should add a TESTING section.

Tests should be included with your software so that it can be tested with a continuous integration platform such as Travis-CI each time modifications are pushed to the repository. It's really handy with pull request too because it can test them before they are merged.

michelf:
The reason those tests are in a separate repository (and do not depend on PHP Unit) is so they can be used by other Markdown implementations. MDTest still needs some tweaking so it can be used without hassle with the new library-style package in the lib branch, but you're right it should get a mention in the readme file. Sorry for the confusion.

PR #127, opened 2013-09-12
added support for the tel: url-scheme

derflocki:
This little patch adds support for the tel: url-scheme

### DIFF
--- a/markdown.php
+++ b/markdown.php
@@ -1447,8 +1447,15 @@ class Markdown_Parser {
 			>
 			}xi',
 			array(&$this, '_doAutoLinks_email_callback'), $text);
+		$text = preg_replace_callback('{<(tel:([^\'">\s]+))>}i',array(&$this, '_doAutoLinks_tel_callback'), $text);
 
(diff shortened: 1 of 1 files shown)


michelf:
Nice idea. I'm not too sure about the syntax though. If you take the email syntax, you can write it directly without a `mailto:` prefix (which is what most people do). I'm not sure if this can be done reliably for a phone number though (various places in the world have various formatting, and this might look too much like a tag).

pjt33:
Could this not be a one-line diff which adds `|tel` to the line

```
$text = preg_replace_callback('{<((https?|ftp|dict):[^\'">\s]+)>}i', 
```

?

michelf:
Eh, that's strange. This pull request should be marked merged, but Github doesn't seem to acknowledge it. It probably has something to do with cherry-picking the commit.

I'm going to close this because it was merged.

@pjt33 The difference with the tel: scheme compared to http and others is that we remove the "tel:" prefix in the link's text content.

PR #141, opened 2014-01-04
command line executable

makryl:
for markdownextra


install
```diff
@@ -0,0 +1,21 @@
...
+
+#
+# Command line executable installer
+# To install run: sudo ./install
+# Usage: markdownextra input1.md [input2.md ...] > output.html
+#
+
+$dir = dirname(__FILE__);
+
+$code = "#!/usr/bin/php\n";
```
sergeyklay:
``` sh
#!/usr/bin/env php
```

Use `env` to find where PHP is installed: it might be elsewhere in the `$PATH`, such as `/usr/local/bin`.

install
```diff
@@ -0,0 +1,21 @@
...
+$code .= str_replace('?>', '', file_get_contents($dir . '/Michelf/MarkdownInterface.php'));
+$code .= str_replace(array('?>', '<?php', 'namespace Michelf;'), '', file_get_contents($dir . '/Michelf/Markdown.php'));
+$code .= str_replace(array('?>', '<?php', 'namespace Michelf;'), '', file_get_contents($dir . '/Michelf/MarkdownExtra.php'));
+$code .= 'for ($i = 1; $i < $argc; $i++) echo MarkdownExtra::defaultTransform(file_get_contents($argv[$i]));';
+
+$filename = '/usr/bin/markdownextra';
```
sergeyklay: `/usr/local/bin/markdownextra`

makryl:
thanks, @sergeyklay

michelf:
So with this we get:
1. a command line tool, nice
2. installed in /usr/local/bin ... and about Windows users? what if you want to generate the tool but not install it? what if you want to install it elsewhere, not as root? 
3. for markdownextra ... what about plain markdown?
4. concatenated in one file with no dependency ... why? 

I'm concerned that this script might break without me noticing when I make updates, for instance if I add a file. I'm also concerned about the example it sets: it shows that "concatenate everything in one file" is okay. Others will necessarily copy it in various forms (to install things the way _they_ want) and complain to me when it breaks.

sergeyklay:
1. --
2. [Windows port of the famous GNU bash](http://win-bash.sourceforge.net). 

@michelf Yes you're right. Script shouldn't be installed. It was easier to use it directly without installation. Eg.

``` sh
$ chmod o+x markdownextra
$ ./markdownextra input1.md [input2.md ...] > output.html
```

or for example Windows users can use it (If user not set `bash.exe` and `markdownextra.sh` into the `PATH` variable)

```
c:\path\to\win\port\of\bash.exe c:\path\to\markdownextra.sh c:\path\to\input1.md [c:\path\to\input2.md ...] > c:\path\to\output.html
```

makryl:
Yes, I will rewrite it with require_once 'Michelf/... , it will work if library exist in same dir or in global include path. Placing script in */bin/ directory is task of deb/rpm package managers.
For windows, .bat file would be OK, no need of "bash.exe"

realityking:
In the day and age of composer, I wonder why this should be done?

robbieaverill:
I agree with @realityking - this surely could be installed globally with composer, then the system bin/path/etc is up to composer to configure, not this repo.

PR #143, opened 2014-01-17
Simplified strong/emphasis regexps

JoshyPHP:
I'm doing research on the Markdown syntax and I was looking through your Markdown library to figure out what was the exact rules with regards to strong/emphasis. While reading the corresponding regular expressions I noticed they were a bit more complicated than necessary.

For instance, the assertion `(?<=\S|^)` means "preceded by not-a-space, or nothing at all". What it really means is "not preceded by a space" which can be expressed as `(?<!\s)`. The same applies to lookahead assertions. Then, once those assertions are simplified they can be merged with compatible assertions around them.

```
(?<=\S|^) becomes (?<!\s)
(?=\S|$)  becomes (?!\s)
(?<!\s)(?<!\*) becomes (?<![\s*])
(?<!\s)(?<!_)  becomes (?<![\s_])
(?!\s)(?![\.,:;]\s) becomes (?![\.,:;]?\s)
```


michelf:
If I apply those changes all the tests are ok which is to be expected as they are equivalent. But my benchmarks show it now runs slightly slower (by 13% for Markdown and 11% for MarkdownExtra respectively).

So I'm not too inclined to apply the changes, but I'm intrigued.

JoshyPHP:
Can you post a gist of your benchmark script please?

michelf:
I'm using [MDTest](https://github.com/michelf/mdtest/).

I made a testsuite where I took all the pages of the [TextMate manual](https://github.com/textmate/manual/tree/master/pages) and concatenated them into one file (All.text). Then I created a second file where the whole manual is repeated twice (Double.text). Those percentages came from parsing this second file. (Note: I probably have an older version of the manual, but I expect the results to be pretty similar.)

In case you're wondering, the reason for those two files is to check that parsing time grows as linearly as possible relative to the size of the document, loosely checking for issues related to algorithmic complexity.

JoshyPHP:
I don't have MDTest setup so here's what I used:

``` php
include '/s9e/php-markdown/Michelf/MarkdownInterface.php';
include '/s9e/php-markdown/Michelf/Markdown.php';

$s=microtime(1);
$html = Michelf\Markdown::defaultTransform(file_get_contents('/tmp/all.md'));
$e=microtime(1);
printf("%.6f\n",$e-$s);

var_dump(crc32($html));
```

`all.md` is the result of `cat *.md > /tmp/all.md` executed in TextMate's manual dir. I've just run this script in CLI (no opcode cache) and via CGI (with OpCache) and I got the same results with or without this PR. That is, the times expressed via `microtime()` as well as the time per request from ApacheBench (run with concurrency 1) were constantly equivalent/within normal deviation. That's consistent with the results I got when I created this PR. I'm currently using PCRE 8.35. I don't know what I was running back then. The times were the same with PHP 5.5.12 and PHP 5.4.28. I also tried HHVM 3.1-dev in CLI, with similar results (that is, no difference between branches.) I'm on Linux.

I guess we must be using different setups.

michelf:
> I guess we must be using different setups.

I'm on OS X 10.9.2, with PHP 5.4.24. Not sure about the PCRE version although it's probably an earlier version than what you have (Apple's PHP bundles rarely are bleeding edge). The PCRE version is most likely the reason we have a difference. That, or maybe I'm not testing it properly. I'll try your benchmark script.

In any case, if the parsing time isn't improved by merging this I'm not sure I should merge it. I admit it's a little cleaner, but I'd rather not needlessly tweak what works already in case some version of PCRE somewhere doesn't optimize things the same way.

michelf:
Whatever I try now I seem to get pretty much the same results for both versions. I suspect something was running in the background when I ran my tests earlier which caused the difference. Thank you for challenging my findings when they looked wrong.

Hum, maybe I should merge those changes then...

JoshyPHP:
That's great news. Originally I thought you were hitting a bad edge case on an old version of PCRE, or perhaps we were benchmarking with some very different inputs, but if you can reproduce on your side that's a relief. :)

PR #148, opened 2014-02-07
Add $this->default_code_class configuration option for fenced code blocks

jackmcdade:
Would love to be able to set a default class on all my fenced code blocks. Most, if not all, syntax highlighters require a class or two, so this would reduce repetitive markup or a brittle javascript dependency.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -1553,6 +1553,10 @@ abstract class _MarkdownExtra_TmpImpl extends \Michelf\Markdown {
 
 	# Optional class prefix for fenced code block.
 	public $code_class_prefix = "";
+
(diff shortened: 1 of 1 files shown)


michelf:
Sorry for taking this long to respond.

In what way is `$this->code_class_prefix` not fulfilling this need? You can put a space in `code_class_prefix` and then you'll get tow classes if you write this:

```
~~~ .php
<?php echo "Hello world"; ?>
~~~
```

Or do you mean you don't want to have to write the language's name?

jackmcdade:
Right, I didn't want to have the language name prefixed. The space thing will work more or less as a workaround though. Carry on!

jackmcdade:
I no longer need this. Not that it was at risk of getting merged anyway. :)

PR #151, opened 2014-02-16
Add 'figure' to list of block-level elements in _MarkdownExtra_TmpImpl

fredleblanc:
`figcaption` is in the list of block-level elements, but `figure` is missing. This adds that in.


michelf:
Good catch. I'll merge this as soon as I have a minute to run the unit tests on it.

PR #153, opened 2014-02-28
Add ID and class attribute support to tables

kburton:
Added support for {#id .class} attributes to tables when attributes are appended to the header line.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -2512,21 +2512,23 @@ abstract class _MarkdownExtra_TmpImpl extends \Michelf\Markdown {
 		#
 		$text = preg_replace_callback('
 			{
-				^							# Start of a line
-				[ ]{0,'.$less_than_tab.'}	# Allowed whitespace.
-				[|]							# Optional leading pipe (present)
-				(.+) \n						# $1: Header row (at least one pipe)
+				^										# Start of a line
+				[ ]{0,'.$less_than_tab.'}				# Allowed whitespace.
+				[|]										# Optional leading pipe (present)
+				(.+?)									# $1: Header row (at least one pipe)
+				(?:'.$this->id_class_attr_catch_re.')?	# $2: Extra table attributes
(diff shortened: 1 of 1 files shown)


kburton:
I've implemented basic table attribute support, primarily so I could add the .table class (and others such as .table-striped) for Bootstrap 3 themed tables. This is used as follows:

```
Testing | Basic | Table {.table-class}
--------|-------|------
a       | b     | c
```

I could fairly easily implement a more powerful solution to allow the following if it would be useful:

```
Testing | Basic | Table {.th-row-class}
--------|-------|------ {.table-class}
a       | b     | c     {.td-row-class}
d       | e     | f     {.td-row-class-2}
```

or with the location of .table-class and .th-row-class switched.

markseuffert:
+1 like. ID/class for the whole table would be nice, to format layout via CSS.

michelf:
I'd rather have the attribute braces go before or after the table, but I can't decide between the two. It seems to me that putting the attribute list just after a row would take too much horizontal space: tables often take much horizontal space and with line wrapping taking more than what you have is inconvenient.

I've also been looking forward adding an optional caption to the table. Right now, I'm leaning towards something that looks like this right now:

```
[ Table 1: Insightful Data ] {.caption-class}
Testing | Basic | Table
--------|-------|------
a       | b     | c
{.table-class}
```

kburton:
I can see what you're saying, but I think that having attributes on separate lines before and after the table obscures the structure. I suppose that's going to be an inevitable consequence of trying to add attribute support to more HTML elements.

jlj:
Same need for me of adding a .table class at the table level for bootstrap.

I was inclined to add the class/id attributes after a tailing '|' of the header row, so as not to confuse it with internal annotation for the table cell, like

```
Testing | Basic | Table |{.table-class}
--------|-------|------
a       | b     | c
```

But the optional caption way would be fine as well. :)

PR #154, opened 2014-02-28
Add ID and class attribute support to tables, header rows and data rows

kburton:
Added support for {#id .class} attributes to full tables, header rows and data rows.


kburton:
I've implemented table attribute support for the full table and header and data rows:

```
Testing | Basic | Table {.table-class}
--------|-------|------ {.th-row-class}
a       | b     | c     {.td-row-class}
d       | e     | f     {.td-row-class-2}
```

I decided to keep the table attributes alongside the header row for consistency with the previous version. The header row attributes are entered on the underline row.

Feel free to merge whichever of these is most appropriate.

PR #155, opened 2014-03-01
Added syntax highlighting to README

demaisj:
Added this syntax highlighting for better reading of the code snippets.


michelf:
I'd rather keep Readme.md compatible with the plain Markdown parser, which has no support for fenced code blocks.

Issue #85, opened 2013-04-09
Ability to create custom filters (e.g. link filter)


xPaw:
(from issue #82)

aoloe:
i'm interested in working on this.
mostly to allow something like https://pypi.python.org/pypi/markowik 's image-baseurl.

my goal is to add a prefix to all local url, both for images and for links.
do you think that it can be solved with filters?

do you already have an idea, how and where custom filters should be implemented?

otherwise, i can imagine that implementing it in a similar way as `$parser->fn_id_prefix` would be a simpler fix for me...
i would go for `url_href_prefix` (for all local paths), `url_a_prefix` (only for links), and `url_img_prefix` (only for images).

michelf:
The way I see it, there should be a `url_filter_func` configuration variable that contains something callable. It should contain a function name string, an array with an object and a method name, an anonymous function, or a closure (available with PHP 5.4). Whenever the parser has to emit some HTML that contains a URL, it calls `url_filter_func` this way:

```
$url = call_user_func($this->url_filter_func, $url);
```

to get the URL to emit in the HTML output. The default value of `url_filter_func` would be a reference to a function that checks for a `mailto:` scheme and obfuscate it with entity encoding (as discussed in #171). Someone who wants to change the default behaviour can set `url_filter_func` with a function that does whatever he wants (adding a prefix, checking for XSS, ban domains, etc.).

aoloe:
looks good!

thanks!

personally, i've spent my time checking what are my "real" needs and implemented a class wrapped around php-markdown.!
what i noticed, is that i need the option for different url transformation for local images and local urls.

i skimmed through your patch and it looks really straight forwards (which is a further compliment to you! :-): if at the end of the current project i'm still convinced that i need that separation, i'll make a patch and pull request (it should not be more than a dozen of lines of code...).

markseuffert:
It looks promising, thanks!

I added the new custom filter to our CMS. I am no longer able to sneak in JavaScript into the generated output. Here's what I did: Enabled `no_markup` and `no_entities`, filtered all URLs via `url_filter_func` which checks for regular expression `/^javascript:/i`. Then we tried to inject HTML code with obfuscation and random data. So far it holds, but hard to tell how well it works. [Here's are some test cases](http://demo.datenstrom.se/tests/xss-filter).

The question is: is this enough to prevent XSS?

michelf:
> The question is: is this enough to prevent XSS?

[...]

PR #156, opened 2014-03-02
Allow strong filtering of linked URLs

iamcal:
Currently, apps using this library for turning user-entered Markdown into HTML are vulnerable to XSS attacks via allowing javascript (and others) as the protocol for links. For example:

```
[CLICK ME](javascript:alert(document.cookie))
```

This PR passes all link URLs through the `filterUrl()` method. By default, this function will allow through URls using a whitelist of protocols, or local/relative URLs without a protocol. There is some heavy decoding involved to make sure URLs like `javascript%3Aalert(document.cookie)` don't slip through.

This change breaks two tests in the current suite ("Links, inline style" & "Quotes in attributes") since they both involve putting broken URLs inside links. Changing the test to have URLs prefixed with `http://` makes these test pass correctly.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -59,6 +59,9 @@ class Markdown implements MarkdownInterface {
 	public $predef_urls = array();
 	public $predef_titles = array();
 
+	# Protocols allows for link targets
+	public $links_protocols = array('http', 'https', 'ftp', 'mailto');
+	public $links_relative = true;
 
 	### Parser Implementation ###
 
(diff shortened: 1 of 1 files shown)


michelf:
By default, PHP Markdown lets HTML snippets pass through unchanged. This is part of the design of Markdown. Markdown is not designed for dealing with unfiltered user input or a substitute for an XSS filter.

Assuming we add one, I don't think an URL filtering system should be active by default, because it'd break many current use cases (starting with my own website) where the input is trusted with relative links as well as HTML snippets.

But the question is: is this enough to prevent XSS? Assuming I wanted the feature, I wouldn't accept anything less than a complete solution, otherwise it'll lead to people thinking they're safe when they're not. Before we can pretend we can filter XSS properly, we should make sure we really do, with some auditing ideally. A glaring omission I see immediately in your patch is the filtering of image URLs. I wouldn't be surprised there are other ways to leak javascript I haven't thought about.

My recommendation is to run an external XSS filter on Markdown's output. This way you know that whatever weird bug the parser has you're still safe. Bolting XSS filtering in a parser that does all those other unrelated complicated string manipulations is a recipe for security holes. I believe it's better to keep XSS filtering as a separate step.

On the subject:
http://michelf.ca/blog/2010/markdown-and-xss/

iamcal:
All good points!

I'm using Markdown with markup and entities disabled, so regular HTML is not preserved. I can certainly understand not wanting to get into having to deal with all the possible transform exploits as security vulnerabilities rather than just plain old bugs.

Worth noting that the patch wont break your current usage - it just breaks some non real-world tests in the suite:

```
url://with spaces
/&quot;style=&quot;color:red
/'style='color:red
```

markseuffert:
I want to add my interest. 

An external XSS filter like [HTML purifier](https://github.com/ezyang/htmlpurifier) is big and slow. I wonder what would happen if you [switch off HTML code](https://michelf.ca/blog/2007/php-markdown-no-markup/) and apply a strong filter to all Markdown generated links (anywhere JavaScript can leak trough). Like Michel said, it must be a complete solution, not just pretending to do the job.

Is this something `url_filter_func` from issue #85 can be used for?

michelf:
@markseu The new `url_filter_func` should allow transforming suspicious URLs into innocuous ones. Feel free to experiment.

PR #160, opened 2014-03-28
Filter potential XSS attack in links

gholol:
It Removes potential XSS attack in links, now only http:// https:// and ftp:// links are allowed. Filter is case insensitive.

```
[XSS Vulnerability](javascript:alert('xss'))
```

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -618,6 +618,10 @@ class Markdown implements MarkdownInterface {
 		$title			=& $matches[7];
 
 		$url = $this->encodeAttribute($url);
+		
+		if(!preg_match('#^\s*(https?|ftp)://#i', $url))
+			$url = '';
+			
 
 		$result = "<a href=\"$url\"";
 		if (isset($title)) {


minchinweb:
so does this filter out relative links?

michelf:
I have multiple issues with this pull request.
- First it assumes that everyone wants to filter javascript. I can assure you that this is not the case. Not everyone uses Markdown with user-generated content. This would need to be configurable.
- This change will filter relative links as well as any other URL scheme (tel, ftp, ssh...). While that might be acceptable in a context where you're dealing with user-generated content, it is not acceptable in the general case. This would need to be configurable in a way where you can choose which URL schemes are allowed.
- You forgot to filter images. While it might be stupid to use a `javascript:` URL for an image, it remains an exploitable XSS vector on certain old browsers (IE 6, I think), and there's no reason not to filter that.

In general I'm reluctant to offer a XSS attack blocker because with the current architecture of the parser it's hard to guaranty there's no bug that'll issue something unexpected and pass the filter. I think it's best to use a well-supported XSS filter on Markdown's _output_ to catch deemed suspicious for the context you're working in.

michelf:
I'm closing this. Javascript is allowed by default and I'm not going to change that. But note that you can now attach you own filter callback for URLs with the `url_filter_func` configuration variable if you want to restrict what is allowed as a link.

PR #165, opened 2014-05-14
Remove trailing ?>

donatj:
Its bad form to have ?>'s on files that are just PHP


Frodox:
It's bad to say "its bad" without any arguments ;D

michelf:
@Jecomire Indeed.

@donatj I though I had removed them all already! Apparently this one slipped. Thanks.

Issue #131, opened 2013-09-21
[Feature Request] <cite> in <blockquote>


MightyPork:
You often need to put a cite note into a blockquote.

```
> Something very smart here,
> other line of the wisdom.

> <cite>&#8213; Some Wise Guy</cite>
```

Now while this works, it's nowhere as neat and readable as this

```
> Something very smart here,
> other line of the wisdom.
> -- Some Wise Guy
```

---

I am using this to do the trick

```
"#^>\s*--\s*(.*)$#im" => "\n> <cite>$1</cite>\n",
```

along with a bit of CSS

```
blockquote cite {
    display: block;
}

blockquote cite:before {
    content: "\2014 \2009";
}
```

How about adding a support for the cite note into MarkdownExtra? Simply the last `> -- Something` will turn into `<cite>` outside the last paragraph, but still inside the blockquote.

ethantw:
This is wrong. You shall not put the author's name into `<cite>`. The semantic element is made to cite the source of the quotes, e.g. the title of a book, a website, or the name of an article, etc.

Below is the example,

```
<blockquote>
<p>Something very smart here, <br>
other line of the wisdom.</p>

<footer>
— Some Wise Guy, <cite>Some Wise Masterpiece</cite>
</footer>
</blockquote>
```

gizmecano:
The use of this element is indeed not so simple (see http://html5doctor.com/blockquote-q-cite/#cite or http://devdocs.io/html/cite). I think that elements `<cite>` and `<q>` are probably one of the main limitations of the Markdown language.

gizmecano:
To add to the confusion about these elements, the specifications seem to have been updated recently, as described in this recent article (http://html5doctor.com/cite-and-blockquote-reloaded/). However, I doubt that this can change anything for this issue...

MightyPork:
Well, I didn't really expect you conservative guys to change anything, but this last definition pretty much matches my use of the tags, doesn't it? Finally the syntax can be used in a way that makes some sense!

vagari:
I'm still hoping for some HTML5 in Markdown at some point. Maybe with an option to turn it on. There are a handful, like figure, that would be nice additions. The HTML5doctor examples make use of footer. [A List Apart uses figure](http://alistapart.com/blog/post/more-thoughts-about-blockquotes-than-are-strictly-required).

michelf:
[...]

PR #167, opened 2014-07-03
Add blockquote footer syntax for citations.

jacktonkin:
Implements the proposed syntax for blockquote footers from #131


michelf:
Thank you. It looks good.

It'd be helpful if some tests were added to [MDTest](https://github.com/michelf/MDTest/) with a few cases that should and should not trigger the feature.

PR #171, opened 2014-08-03
Entity encode mailto anchors

sorbits:
Previously the following was left unencoded: `[send feedback](mailto:feedback@example.org)`.

This commit introduces a public `Markdown::entityEncodeText` (static) function since we need to call this from an anonymous function and probably want to keep compatibility with older PHP versions (which lack closures). Though if the public function is undesired, we can drop the anonymous function and pass `array($this, «function name»)` as callback, which seems to be the current convention.

I just wanted to ensure you actually want this feature before spending further time on it.

There does seem to be some potential for factoring out link creation into a single function accepting URL, text, and optional title, this function would then check for `mailto:` URL prefix and do appropriate processing of text and title as well.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -594,6 +594,7 @@ class Markdown implements MarkdownInterface {
 		if (isset($this->urls[$link_id])) {
 			$url = $this->urls[$link_id];
 			$url = $this->encodeAttribute($url);
+			$url = $this->encodeMailToAddressInURL($url);
 			
 			$result = "<a href=\"$url\"";
 			if ( isset( $this->titles[$link_id] ) ) {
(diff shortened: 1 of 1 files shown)


michelf:
I'd rather continue to use use the array callback method.

It'd be really nice if all URLs were processed by the same function at some point. There has been some requests about adding a hook that could be used to filter links, it'd be a step towards that.

As for encoding emails in all links, I feel it's a good idea.

sorbits:
Commit updated to use an external callback function (and does not introduce any public functions).

Michelf/Markdown.php
```diff
@@ -1345,29 +1356,30 @@ protected function encodeEmailAddress($addr) {
...
-		$chars = preg_split('/(?<!^)(?!$)/', $addr);
-		$seed = (int)abs(crc32($addr) / strlen($addr)); # Deterministic seed.
-		
+		$mailto = $this->entityObfuscateText("mailto:");
+		$email  = $this->entityObfuscateText($addr);
+		return "<a href=\"{$mailto}{$email}\">{$email}</a>";
+	}
+
+
+	private function entityObfuscateText($text) {
```
michelf, line 1365: I think this could be `protected`, like other functions. Just don't make it `public`.

michelf:
This was implemented by a8c56ec.

PR #180, opened 2014-10-21
Updates README. Adds Composer Install Instructions

erikthedeveloper:
This package is on Packagist (https://packagist.org/packages/michelf/php-markdown). 
This package promotes PSR autoloading. 
~~Hence, this package should be installed via Composer and promote the autoloading capabilities provided by Composer.~~
This makes utilizing Composer and its autoloading directive a pleasure as a user/developer.
- http://www.phptherightway.com/#dependency_management
- http://blog.doh.ms/2014/10/13/installing-composer-packages/

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -53,12 +53,12 @@ Using [Composer](https://getcomposer.org/) (the preferred method) simply run
 composer require michelf/php-markdown
 ```
 
-and the package will be installed into your `./vendor` directory.
+and the package will be installed into your `./vendor` directory and available for use. *Make sure to `require` Composer's `vendor/autoload.php` to enable autoloading throughout your project.*
 
 Usage
 -----
 
-This library package is meant to be used with class autoloading. *This is taken care of automatically if you installed via Composer.* For autoloading 
+This library package is meant to be used with class autoloading. For autoloading 
 to work, your project needs have setup a PSR-0-compatible autoloader. See the 
 included Readme.php file for a minimal autoloader setup. (If you cannot use 
 autoloading, see below.)
--- a/Readme.md
+++ b/Readme.md
@@ -43,11 +43,22 @@ Before PHP 5.3.7, pcre.backtrack_limit defaults to 100 000, which is too small
 in many situations. You might need to set it to higher values. Later PHP 
 releases defaults to 1 000 000, which is usually fine.
 
+Installation
+------------
+
+Using [Composer](https://getcomposer.org/) (the preferred method) simply run
+
+```
+# This will ensure you get the latest stable version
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -43,11 +43,22 @@ Before PHP 5.3.7, pcre.backtrack_limit defaults to 100 000, which is too small
...
+composer require michelf/php-markdown
+```
+
+and the package will be installed into your `./vendor` directory.
 
 Usage
 -----
 
-This library package is meant to be used with class autoloading. For autoloading 
+This library package is meant to be used with class autoloading. *This is taken care of automatically if you installed via Composer.* For autoloading
```
michelf: Isn't that an exageration? Don't you have to `require 'vendor/autoload.php'` for autoloading via Composer to work? It [says so in the documentation](https://getcomposer.org/doc/00-intro.md#autoloading) at least.

erikthedeveloper: Yeah, you are right (I must have spaced it). I guess this is assuming knowledge of composer and how its autoloading works, which won't be the case for composer newcomers. hmm.. Could use a link to the composer autoloading docs or maybe add a line like you suggested to explain that you have to

"add the `require 'vendor/autoload.php'` somewhere."

in order for it to actually be usable. 

But yes. Could use some clarification for composer newcomers and as use as the singular dependency.

michelf:
> Hence, this package should be installed via Composer and promote the autoloading capabilities provided by Composer.

I'm not sure why you come to that conclusion. While I want to make it easy to use with Composer, it is not my intention to push a particular packaging system to my users. Though I do agree the Readme could be clearer about how to install and use with Composer. And I like your installation instructions.

Readme.md
michelf: Could add a line like this one: "If you are using Composer, it setups autoloading for you, see below.". Then add a "Composer" section just above the "Usage without an autoloader" section. I'd also put the installation instructions there so that everything about composer is at the same place.

Readme.md
```diff
@@ -43,11 +43,22 @@ Before PHP 5.3.7, pcre.backtrack_limit defaults to 100 000, which is too small
...
 
+Installation
+------------
+
+Using [Composer](https://getcomposer.org/) (the preferred method) simply run
+
+```
+# This will ensure you get the latest stable version
+composer require michelf/php-markdown
+```
```
michelf: Please use an indented code block to keep this document compatible with the plain Markdown parser. Fenced code blocks are a feature of the MarkdownExtra parser (and Github), but this document isn't targeted solely for those.

erikthedeveloper: Good point. Will keep that in mind for future.

PR #185, opened 2014-11-20
Auto loading id's

zquintana:
For auto loading id's in heading tags. Very useful for speeding up dev of things like documentations.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -1682,14 +1682,15 @@ abstract class _MarkdownExtra_TmpImpl extends \Michelf\Markdown {
 	# Expression to use when parsing in a context when no capture is desired
 	protected $id_class_attr_nocatch_re = '\{(?:[ ]*[#.][-_:a-zA-Z0-9]+){1,}[ ]*\}';
 
-	protected function doExtraAttributes($tag_name, $attr) {
+	protected function doExtraAttributes($tag_name, $attr, $text='') {
 	#
 	# Parse attributes caught by the $this->id_class_attr_catch_re expression
 	# and return the HTML-formatted list of attributes.
 	#
 	# Currently supported attributes are .class and #id.
 	#
-		if (empty($attr)) return "";
+		$is_heading = preg_match('/^(h|H)[1-6]{1}/', $tag_name) ? true : false;
+		if (!$is_heading && empty($attr)) return "";
 		
 		# Split on components
 		preg_match_all('/[#.][-_:a-zA-Z0-9]+/', $attr, $matches);
@@ -1704,6 +1705,10 @@ abstract class _MarkdownExtra_TmpImpl extends \Michelf\Markdown {
 			} else if ($element{0} == '#') {
 				if ($id === false) $id = substr($element, 1);
 			}
+		}
+
+		if ($is_heading && $id === false) {
+			$id = preg_replace('/\W/', '-', strtolower($text));
 		}
 
 		# compose attributes as string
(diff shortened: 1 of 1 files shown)


michelf:
I'm not sure there's a good solution to this problem, but it seems to me that `'/\W/` is slightly lacking as a regex to remove undesirable characters for the `id` attribute: it's going to give you garbage results in many languages. Even in English, what does it do with "Résumé"? Now extrapolate to Japanese or Hebrew.

So I think this needs some work. If we release the feature like this and later update the filtering algorithm, it'll make a lot of `id` attributes unstable and risk creating broken links. I think we need to try to get it right the first time. But also I don't think all issues can be solved so perhaps it should only be enabled through a configuration variable.

Or perhaps the configuration variable could be a filter function? Something like the newly added URL filter... some room for thoughts. Any idea what the other Markdown implementations do with that?

Also, what about documents with two headers sharing the same textual content?

zquintana:
I didn't think about the first potential problem, but it is definitely a
potential problem. I'm not sure what other languages do, I was using both
Maruku and Kramdown before, but had to move to a php solution. Those are
both ruby. Those solutions were auto inputting the id's based on the header
text which I got used to. It was useful for auto creating table of contents
section (which was another feature of both those Ruby libs). Here's an
example we are using right now with this
https://licensing.glendaledesigns.com/docs/gd-variables. I ran a quick test
and it looks like Kramdown is removing those accented chars and it's
indexing the id names when multiples are present. So in you're
example Résumé turns into rsum first name and rsum-1 on the second. Here is
a snippet of the doc:

# Résumé

# Résumé

I can add code to that effect if you would prefer to see that. What do you
think?

As for other idea is to add a compile flag to turn this feature on/off.
Although I'd definitely would like to see it on by default, but I can work
with it either way.

zquintana:
Here is some code to handle the above issues:

``` php
# Markdown.php

# Line: 1569
# Used to keep track of auto generated ids and their counts
private $_auto_ids = array();

# Line: 1686
if ($is_heading && $id === false) {
    $id     = preg_replace('/\s/', '-', $text);
    $id     = preg_replace("/[^a-zA-Z0-9_-]/", '', strtolower($id));

    if (isset($this->_auto_ids[$id])) {
        $index  = $this->_auto_ids[$id];
        $this->_auto_ids[$id]++;

        $id     = "$id-$index";
    }

    $this->_auto_ids[$id] = 1;
}
```

Tested this markdown

``` md
# Test

# Test

Test 2
---------

# Résumé

# Résumé

# Résumé
```

Resulted in

``` html
<h1 id="test-1">Test</h1>

<h1 id="test-2">Test</h1>

<h2 id="test">Test</h2>

<h1 id="rsum">Résumé</h1>

<h1 id="rsum-1">Résumé</h1>

<h1 id="rsum-2">Résumé</h1>
```

Any thoughts?

michelf:
I really don't like turning "résumé" into "rsum". Couldn't we use [Transliterator::transliterate](http://php.net/manual/en/transliterator.transliterate.php)? That'll limit the feature to PHP 5.4, unless we allow custom transliterator implementations through a configuration variable.

There's also the question of markup inside of the header:

```
# <em>Résumé</em>

# *Résumé*
```

Look at [the funny results in Babelmark 2](http://johnmacfarlane.net/babelmark2/?normalize=1&text=%23+%3Cem%3ERésumé%3C%2Fem%3E%0A%0A%23+*Résumé*). Anyway, I think the markup should be stripped before passing it to the transliteration function.

Are you sure your input in your last comment correctly matches the output?

michelf:
Another consideration: when displaying multiple documents on the same page (on a blog for instance), you'll want the ability to set a prefix to all ids on that page (especially the generated ones). There's already a setting for that for footnote ids. Perhaps we need another for headers, or perhaps we should use the same prefix, I'm not too sure.

michelf:
Pull request #196  has been merged instead of this one. There's now a `header_id_func` configuration variable where you can set a function that'll be used to create an ID from the header.

Issue #93, opened 2013-04-28
Nofollow links


xPaw:
Would be nice to have an option to make all links `rel="nofollow"`.

funivan:
it will be nice to add any attribute to any block
For example:

```
# h1 title # [::attibute]
## h2 title ## [::attibute]

[attribute::][data-tip="test"]
```

And it convert to 

```
<h1 data-tip="test">h1 title</h1>
<h2 data-tip="test">h2 title</h2>
```

michelf:
If I offer this it'll be a setting on the parser, not something within the syntax. I'm seriously considering offering a callback in the API to allow altering emitted link URLs and attributes, which would make this rather easy to do.

funivan:
Callback will be great solution =)

attiks:
I need something similar to be able to add a data- attribute to image tags, so I can post process the output to turn some images into picture.

Is there any progress on this issue?

I saw there's already support for id and class using `{#id .picture}` maybe we can do something like `{#id .picture :data-group=test}` which gets translated into `<img data-group=test>`

michelf:
> Is there any progress on this issue?

Not much. The time can afford to put on PHP Markdown is rather limited these days.

> I saw there's already support for id and class using `{#id .picture}` maybe we can do something like `{#id .picture :data-group=test}` which gets translated into `<img data-group=test>`

If you want to give it a shot, look at the `doExtraAttributes` function, as well as the `id_class_attr_catch_re` and `id_class_attr_nocatch_re` regular expressions.

jeffmcneill:
Google is strongly recommending href attributes. Some solution is needed. https://developers.google.com/search/docs/advanced/guidelines/qualify-outbound-links

michelf:
This should have been closed earlier. You can write `[link](url){rel=nofollow}` using the MarkdownExtra parser since 2014.

michelf:
Oh, well... maybe I closed it too soon. This is about adding the attribute to all links. It should probably work using something like the `url_filter_func` configuration but for the `rel` attribute.

IzzySoft:
While on it, and if it's not asked too much: I'd love to see an option to have all external links `rel=nofollow` by default and just mark exceptions (maybe like `[link](url){rel=}` to unset the attribute). Would that be possible (and feasible)?

PR #186, opened 2014-12-07
Add support for any attribute

attiks:
This PR adds support for any attribute as discussed in #93 

Examples:
- `{:data-group=test}`
- `{:rel=nofollow}`

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -1678,9 +1678,9 @@ abstract class _MarkdownExtra_TmpImpl extends \Michelf\Markdown {
 	### Extra Attribute Parser ###
 
 	# Expression to use to catch attributes (includes the braces)
-	protected $id_class_attr_catch_re = '\{((?:[ ]*[#.][-_:a-zA-Z0-9]+){1,})[ ]*\}';
+	protected $id_class_attr_catch_re = '\{((?:[ ]*[#.:][-_:a-zA-Z0-9=]+){1,})[ ]*\}';
 	# Expression to use when parsing in a context when no capture is desired
-	protected $id_class_attr_nocatch_re = '\{(?:[ ]*[#.][-_:a-zA-Z0-9]+){1,}[ ]*\}';
+	protected $id_class_attr_nocatch_re = '\{(?:[ ]*[#.:][-_:a-zA-Z0-9=]+){1,}[ ]*\}';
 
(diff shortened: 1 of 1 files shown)


michelf:
Nice.

One thing I'll have to think about is security considerations. Currently someone can enable the no-markup mode and add a callback for changing URLs to filter javascript and other things that could be exploited in user-generated content. This PR pierces hole in those measures. Perhaps this feature should simply be disabled when the no-markup flag is set.

Also, is there a reason for wanting a colon as a prefix? Why not simply accept `key=val`?

attiks:
I added `!$this->no_markup &&`

Mainly to be consistent with the id and class attributes, since they are using a prefix as well, but I don't mind changing it.

michelf:
I'd rather have `key=val`. Especially since there's already a precedent of doing it like this in another implementation (maruku).

attiks:
I removed the ':' and just checked for a-z as the first character

Michelf/Markdown.php
```diff
@@ -1692,17 +1692,21 @@ protected function doExtraAttributes($tag_name, $attr) {
...
+		$attributes = array();
 		$id = false;
 		foreach ($elements as $element) {
 			if ($element{0} == '.') {
 				$classes[] = substr($element, 1);
 			} else if ($element{0} == '#') {
 				if ($id === false) $id = substr($element, 1);
+			} else if (strpos($element, '=') > 0) {
+				$parts = explode('=', $element);
+				$attributes[] = $parts[0] . '="' . $parts[1] . '"';
```
michelf, line 1709: Probably should use `explode('=', $element, 2)` since a third equal sign shouldn't cause the attribute value to end. At least that's now HTML parses unquoted attributes.

michelf:
> a-z as the first character

Why not `A-Z` too?

attiks:
Common practice is to write attribute in lowercase only, the spec allows uppercase in theory, but all uppercase characters are - internally - converted to lowercase.

The data- attribute spec demands lowercase only.

michelf:
I agree that from an authoring perspective it's best to use lowercase names. But from the parser's perspective, I'm not sure it's a good idea to ignore attributes blocks with attributes set in uppercase. I guess it's rare enough that we can only bother if someone reports the issue at some point.

Will merge this shortly.

attiks:
Thanks

PR #199, opened 2015-04-14
Subclass of MardownExtra added with Geshi support (syntax highlighting)

mariokonrad:
Hello there!

I have implemented a class MarkdownExtraGeshi (subclass of MarkdownExtra) with basic Geshi support (code syntax highlighing).

It is able to do syntax highlighting for code snippets like this:

``` cpp
int main(int, char **)
{
  // some code
}
```

with a twist: you can select the starting line number of the code, example:

``` cpp:23
std::string s = "Hello World!";
std::cout << s << "\n";
```

Of course, Geshi is needed for this to work, so an additional external dependency. But only for the new class.

Are you interested in a patch like this?

Best regards
Mario

### DIFF
--- a/Michelf/MarkdownExtraGeshi.inc.php
+++ b/Michelf/MarkdownExtraGeshi.inc.php
@@ -1,0 +1,13 @@
+<?php
+
+# Use this file if you cannot use class autoloading. It will include all the
+# files needed for the MarkdownExtraGeshi parser.
+#
+# Take a look at the PSR-0-compatible class autoloading implementation
+# in the Readme.php file if you want a simple autoloader setup.
+
+require_once dirname(__FILE__) . '/MarkdownInterface.php';
+require_once dirname(__FILE__) . '/Markdown.php';
+require_once dirname(__FILE__) . '/MarkdownExtra.php';
+require_once dirname(__FILE__) . '/MarkdownExtraGeshi.php';
+
--- a/Michelf/MarkdownExtraGeshi.php
+++ b/Michelf/MarkdownExtraGeshi.php
@@ -1,0 +1,106 @@
+<?php
+#
+# Markdown Extra Geshi -  A text-to-HTML conversion tool for web writers
+#
+# PHP Markdown Extra Geshi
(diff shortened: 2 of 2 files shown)


michelf:
It's hard to understand what you mean since Github parsed your code block markers. You should probably indent them with four spaces so the markers appear on Github.

One question, regarding the cpp:44 syntax for line numbers, where does it come from?

mariokonrad:
Sorry, didn't think about GitHub eating my examples. New try, normal code block (the same way GitHub parses and highlights code as it seems):

```
~~~ cpp
int main(int, char **)
{
    // some code
}
~~~
```

and with a starting line number:

```
~~~ cpp:44
std::string s = "Hello World!";
std::cout << s << "\n";
~~~
```

The notation "cpp:44" is not derived from GitHubs or Geshis behaviour, it seemed familiar, like error messages from gcc, breakpoint commands from gdb...

michelf:
I don't like that dependency, and I don't like that it has to be a separate class. I would propose this instead: what if there was a configuration variable in the parser letting you specify a filter function for code blocks? Something similar to `url_filter_func` or `header_id_func`. Then you could use Geshi this way:

```
$parser = new \Michelf\MarkdownExtra();
$parser->code_block_content_func = function($code, $lang) {
    $geshi = new \Geshi($code, $lang);
    $codeblock = $geshi->parse_code();
};
$html = $parser->transform($text);
```

As for the line number, I'm not sure I want to have this in MarkdownExtra. But if you want to add `:` to the characters allowed in the language name, I think that would be fine. You could then split and handle it in your custom function.

What do you think?

mariokonrad:
I agree, the additional dependency is not handy, especially because Geshi could be installed anywhere and tinkering with include paths is a mess. Actually (after a second check) the ':' is already allowed in the classname, which makes it even easier.
I have already a working version with your proposal in my forked repo, branch: code_block_content_func. Probably it is easier to just close this pull request and I open another one.

Thanks!

mariokonrad:
Replaced with pull request #200

PR #200, opened 2015-04-16
Optional code block content function added in MarkdownExtra.

mariokonrad:
This adds an optional code block content function to use customized renderers for code blocks (for example Geshi).

### DIFF
--- a/Michelf/MarkdownExtra.php
+++ b/Michelf/MarkdownExtra.php
@@ -1295,10 +1295,10 @@ class MarkdownExtra extends \Michelf\Markdown {
 					(?:~{3,}|`{3,}) # 3 or more tildes/backticks.
 				)
 				[ ]*
-				(?: \.?([-_:a-zA-Z0-9]+) )? # 2: standalone class name
-				[ ]*
-				(?:	'.$this->id_class_attr_catch_re.' )? # 3: Extra attributes
-				
+				# #2: class name (optional) and #3: attributes (optional)
+				(?:
+					\.?([-_:a-zA-Z0-9]+)?[ ]*('.$this->id_class_attr_catch_re.')?
+				)?
 				[ ]* \n # Whitespace and newline following marker.
 				
 				# 4: Content
@@ -1319,7 +1319,7 @@ class MarkdownExtra extends \Michelf\Markdown {
 	protected function _doFencedCodeBlocks_callback($matches) {
 		$classname =& $matches[2];
 		$attrs     =& $matches[3];
-		$codeblock = $matches[4];
+		$codeblock = $matches[5];
 
 		if ($this->code_block_content_func) {
 			$codeblock = call_user_func($this->code_block_content_func, $codeblock, $classname);
--- a/Michelf/MarkdownExtra.php
+++ b/Michelf/MarkdownExtra.php
@@ -1292,11 +1292,10 @@ class MarkdownExtra extends \Michelf\Markdown {
 					(?:~{3,}|`{3,}) # 3 or more tildes/backticks.
 				)
 				[ ]*
-				(?:
-					\.?([-_:a-zA-Z0-9]+) # 2: standalone class name
-				|
-					'.$this->id_class_attr_catch_re.' # 3: Extra attributes
-				)?
+				(?: \.?([-_:a-zA-Z0-9]+) )? # 2: standalone class name
+				[ ]*
+				(?:	'.$this->id_class_attr_catch_re.' )? # 3: Extra attributes
+				
 				[ ]* \n # Whitespace and newline following marker.
 				
 				# 4: Content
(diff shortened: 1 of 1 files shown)


Michelf/MarkdownExtra.php
```diff
@@ -1318,6 +1321,12 @@ protected function _doFencedCodeBlocks_callback($matches) {
 		$classname =& $matches[2];
 		$attrs     =& $matches[3];
 		$codeblock = $matches[4];
+
+		if ($this->code_block_content_func) {
+			$codeblock = call_user_func($this->code_block_content_func, $codeblock, $classname);
+			return "\n\n".$this->hashBlock($codeblock)."\n\n";
+		}
+
```
michelf, line 1332: Hum, I'm not sure I like the way it works, taking charge of emitting the `<pre><code>` and bypassing the extra attribute handling done below. Wouldn't it be enough if the callback only replaced the htmlspecialchars step below instead?

mariokonrad:
As far as I can tell (and tested it with Geshi), there's no difference. The difference does only appear if there are multiple attributes specified for a code block like:

```
~~~ {cpp .something}
// code
~~~
```

because then `$matches` is different and with the proposed patch unusable. It would be indeed difficult to decide which of the "classnames" or part of `$matches` represents the language. The original pull request didn't have this problem, due to the subclass and its different matcher.
However, as long as only one attribute is specified:

```
~~~ cpp
// code
~~~
```

there seems to be no problem.

But to have the _valid_ option to specify multiple attributes that are probably harmful for the custom function, renders it useless...

michelf:
One thing I wish to allow but won't be accepted syntactically at the moment is this:

```
~~~ cpp {.other .things #id}
// code
~~~
```

The one outside the braces is the language, the rest is just some other attributes you want to apply, for instance a `wrong` class to show that some code is proscribed.

mariokonrad:
Sounds reasonable. Would you accept starting line numbers in the following form for the language as well?

```
~~~ cpp:22 { .other .things #id }
// code
~~~
```

michelf:
It seems that the colon is already allowed in the language name. Just have your handler function extract the number from its language parameter and act accordingly. I don't think it makes much sense for PHP Markdown to advertise this is a line number parameter syntax but not do anything with it.

mariokonrad:
Ok, the latest commit supports the syntax above. Both language and attributes are optional. All the following combinations are valid:

```
~~~ { .some .thing }
// ...
~~~

~~~ cpp
// ...
~~~

~~~ cpp { .some .thing #id }
// ...
~~~

~~~
// ...
~~~
```

And the custom code block function only replaces `htmlspecialchars(...)`.

What do you think about the latest commit?

Michelf/MarkdownExtra.php
```diff
@@ -1292,10 +1295,9 @@ protected function doFencedCodeBlocks($text) {
 					(?:~{3,}|`{3,}) # 3 or more tildes/backticks.
 				)
 				[ ]*
+				# #2: class name (optional) and #3: attributes (optional)
 				(?:
-					\.?([-_:a-zA-Z0-9]+) # 2: standalone class name
-				|
-					'.$this->id_class_attr_catch_re.' # 3: Extra attributes
+					\.?([-_:a-zA-Z0-9]+)?[ ]*('.$this->id_class_attr_catch_re.')?
```
michelf: This regex has a two issues:
1. The optional dot before the language name must not be allowed by itself, it must be followed by a language name.
2. `$this->id_class_attr_catch_re` already contains a capture group so you shouldn't wrap it in one. You could wrap it in a non-capture group `(?:)` to make it optional.

So it'll look like this:

```
(?:
    \.?([-_:a-zA-Z0-9]+) # 2: standalone class name
)?
[ ]*
(?:
    '.$this->id_class_attr_catch_re.' # 3: Extra attributes
)?
```

mariokonrad:
Sorry, I'm not that familiar with matchers in PHP. I included your example into the matcher.

michelf:
Sorry for the lack of comments in 9 days. I just need to pass it through MDTest to check it doesn't break anything (which would surprise me) and add a new test case for it. After that I'll merge.

mariokonrad:
Cool! Thanks for the feedback!

markseuffert:
Anything I can do to help? Let me know if you need a test case or something else.

michelf:
I just [added two tests to MDTest](https://github.com/michelf/mdtest/commit/616d1004901be1f2e066be2664a546af89111630) looking like that:

```
~~~~~ .html{.bold}
<b>bold</ b>
~~~~~

~~~~~ .html {#codeid}
<b>bold</b>
~~~~~
```

and they both fail. Here is the result I get:

```
<pre><code class="html">&lt;b&gt;bold&lt;/b&gt;
</code></pre>

<pre><code class="html">&lt;b&gt;bold&lt;/b&gt;
</code></pre>
```

While the regex was updated to support this new form, the code handling the result ignores the special attribute block when there is a standalone class name before it (which was fine before, but not now that they are allowed together).

mariokonrad:
Ok, I have rebased (and unfortunately had to force push) the branch to the latest lib branch. From my basic and humble tests, it appears to work...

michelf:
Looks like it'd work, but I don't like the trick of feeding `doExtraAttributes` with a string to parse: this makes this code's correctness dependent on the syntax of `doExtraAttributes` rather than the interface of the function.

So instead I merged the older version of this pull request with a few fixes of my own. See c9220d1bb3de83c048c083681a1ea5d46c64d913 for my changes.

mariokonrad:
Thanks!

markseuffert:
Thank you guys!

Issue #211, opened 2015-10-09
[feature-request] Strikethrough (del tag) support


kiprobinson:
In GitHub-flavored Markdown, they support strikethrough (`<del>`) using `~~text to strikethrough~~`. Could you add this to Markdown Extra?

Example: here is some ~~deleted~~ text.

Reference: https://help.github.com/articles/github-flavored-markdown/#strikethrough

I went ahead an implemented this. Pull request here: https://github.com/michelf/php-markdown/pull/212

markseu:
+1 Useful feature, would also like to use it

jeffmcneill:
It is time for some kind of strikethrough support in Markdown Extra. The exact syntax does not matter to me. Some kind of syntax is what matters. Why must we resort to html for this when Markdown is supposed to make this easier?

PR #212, opened 2015-10-10
Adding support for ~~strikethrough~~ in MarkdownExtra

kiprobinson:
This adds support for ~~strikethrough~~ syntax (`~~strikethrough~~`) in MarkdownExtra.

Related issue: https://github.com/michelf/php-markdown/issues/211

### DIFF
--- a/Michelf/MarkdownExtra.php
+++ b/Michelf/MarkdownExtra.php
@@ -71,6 +71,7 @@ class MarkdownExtra extends \Michelf\Markdown {
 			);
 		$this->span_gamut += array(
 			"doFootnotes"        => 5,
+			"doStrikethrough"    => 55,
 			"doAbbreviations"    => 70,
 			);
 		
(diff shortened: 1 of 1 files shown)


michelf:
Thank you for this, but I don't intend to merge it. To me that shortcut syntax for `<del>` is not self-explanatory enough to the reader (using `<del>` directly clearer to me). It'd also feel strange to have that without having a shortcut for `<ins>`.

robsonsobral:
@michelf , I tend to disagree simply because I've seen this a lot here in Brazil. On facebook, on twitter... I always thought it's an confusing syntax. Something like `--` is much easier to understand. However, it's becoming a standard for non-techy people to write `~~something this way~~`.

michelf:
Perhaps I should revise this decision some day. But, right now, all I can say is that the syntax doesn't convey anything to me when I read it, and I have yet to see this used in the wild. It has some potential to be confusing with the fenced code block syntax too because the marker is quite similar.

taufik-nurrohman:
What about:

```
--del-- ++ins++
```

michelf:
@tovic The problem with `--del--` is that `--` is already widely in use for em dash (its 's old as the typewriter days). In the middle of a sentence it'll look like a dash to most people, not a strikethrough. You could maybe disambiguate with spacing, but in usage spacing is often inconsistent around dashes between authors so it's probably not a good idea.

robsonsobral:
@tovic , I agree that makes a lot more sense. I suggested it to @michelf years ago by e-mail. But besides the reason he gave, people are already using `~~` and it's easier to embrace a standard than to fight one. Markdown is going to suffer of AFS.

*[AFS]: Androidish Fragmentation Syndrome

Ops. It looks like it is already!

kiprobinson:
"I have yet to see this used in the wild".

Well, GitHub-flavored markdown uses it. That's how I was able to use strikethrough in the pull request. :)

Use it or don't, I don't care... I just implemented it for my own site and thought I'd submit a pull request in case you were interested.

taufik-nurrohman:
Or with braces, like hyperlink (but not realy readable):

```
--[text to delete](@2015-10-13) ++[replacement text](@2015-10-13)
```

tobiastom:
I think `~~` is widely used now (for example on GitHub), and I would like to see this added as well.

MrPetovan:
`~~` is used on Diaspora as well.

michaelbutler:
I agree that using `~~` for strikethrough ~~is~~ seems very common and I personally use it a fair amount (GitHub, Slack). @michelf have you changed your opinion on this?

jeffmcneill:
It is time for some kind of strikethrough support in Markdown Extra. The exact syntax does not matter to me. Some kind of syntax is what matters. Why must we resort to html for this when Markdown is supposed to make this easier?

PR #214, opened 2015-10-16
Prevent iOS from mangling foot-note back-links

okdana:
Safari (and probably other browsers) on iOS automatically converts the '↩' character used for foot-note back-links into [an emoji](https://www.emojibase.com/emoji/21a9/leftwardsarrowwithhook), which is inconsistent with desk-top browsers and (i think) just kind of dumb-looking. iOS can be [told to disable this behaviour](https://paulofierro.com/blog/2013/3/30/disabling-emoji-characters-with-unicode-variants) by appending a Unicode [variant selector](https://en.wikipedia.org/wiki/Variant_form_%28Unicode%29), `U+FE0E`, immediately following the would-be-emoji character. Unfortunately, it's not possible to do this with an `::after` selector in CSS, so the change must be made in the HTML itself. This PR makes that change.

I tested this in the following browsers and it worked as expected:
- Safari on iOS 9
- Firefox 43 on OS X 10.11
- Chrome 45 on OS X 10.11
- Safari 9 on OS X 10.11
- Chrome 44 on Ubuntu 14.04

### DIFF
--- a/Michelf/MarkdownExtra.php
+++ b/Michelf/MarkdownExtra.php
@@ -32,6 +32,12 @@ class MarkdownExtra extends \Michelf\Markdown {
 	public $fn_link_class = "footnote-ref";
 	public $fn_backlink_class = "footnote-backref";
 
+	# Text to be displayed within footnote backlinks. The default is '↩'; the
+	# U+FE0E on the end is a Unicode variant selector used to prevent iOS from
+	# displaying the arrow character as an emoji. Note: This value is inserted
+	# as raw HTML, so dangerous/special characters must be pre-escaped!
+	public $fn_backlink_text = '&#8617;&#xFE0E;';
+
 	# Class name for table cell alignment (%% replaced left/center/right)
 	# For instance: 'go-%%' becomes 'go-left' or 'go-right' or 'go-center'
 	# If empty, the align attribute is used instead of a class name.
@@ -1475,6 +1481,7 @@ class MarkdownExtra extends \Michelf\Markdown {
 				$title = $this->encodeAttribute($title);
 				$attr .= " title=\"$title\"";
 			}
+			$backlink_text = $this->fn_backlink_text;
 			$num = 0;
 			
 			while (!empty($this->footnotes_ordered)) {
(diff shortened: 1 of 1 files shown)


michelf:
Are you trying to confuse me? It's not `U+FE0E` (as in the text above) but `U+FE0F` (as in the code) one has to use to avoid seeing an emoji. I tested the wrong thing a few times before I noticed.

I agree this is a problem that should be fixed. I'll check in a couple more browsers and probably merge this.

michelf:
Many older browsers are showing this incorrectly (displaying the variant selector as a square character). Here is those I found with what I have at hand:
- Safari 5 and less on OS X (so any Safari user on OS X 10.6, 10.7)
- IE 9 and less

I think it's fair to assume that any browser that predates the variant selectors in Unicode are going to display this wrong. So that's probably going to mess things for a lot of Android phones which tend to not be update, and which I think don't have the emoji issue in the first place.

Hum, perhaps I should file a bug with Apple about the emoji and see if they close it as intended behaviour. Do you have a couple of URLs with ↩ showing up as emoji to build a case?

If we make this change, I think we should actually make the backlink marker a configuration variable so it's easy to go back for those who might not find that acceptable.

michelf:
(Oops, I edited my last comment. iOS shouldn't have been in the list since I could not test any older version, unfortunately.)

michelf:
I made a test page to check browser support: https://michelf.ca/docs/test-backlink.html

okdana:
Oh, i was the one who got confused. I made an error in the code; i was too lazy to clone the repo so i just manually transcribed the change from my machine (where i tested it) into GitHub's editor, and i guess i got it wrong. I'm really sorry about that. `U+FE0E` is in fact what fixes it for me.

There are a bunch of pre-existing characters that are marked as emoji in the Unicode standard, including all of the playing-card suits, most arrows, most cross/circle/tick symbols, &c. These are in the dingbats and uncategorised ranges on [this list](http://apps.timwhitlock.info/emoji/tables/unicode) for example. iOS has been displaying these as emoji for quite some time, i think since at least iOS 6.

I rebased with the fix for my error, though if you are concerned about older browsers i agree it's not sufficient as-is. I can make it configurable if you like — should i just add it as a public class property? `fn_backlink_text` maybe? Any other concerns?

michelf:
I think the two of us got confused about U+FE0F at some point. But that does't matter.

`fn_backlink_text` would be nice, with a default value of `&#8617;&#xFE0E;` and a comment next to it explaining that the `&#xFE0E;` is to select the non-emoji variant on iOS (because people will wonder).

okdana:
Alright, i've rebased again per your feedback. Is this OK?

okdana:
Thank you!

michelf:
Note: I made a little change in that `fn_backlink_text` is now called `fn_backlink_html` to clarify that it is indeed HTML content.

Also updated the version history in Readme.md. I credited "Dana" for the code (took the name from your profile page). If you want me to put a full name, let me know.

okdana:
That's great, thank you. I normally spell it 'dana', but that's OK too. Thanks again!

PR #222, opened 2016-02-04
Optional code span function added

styxit:
Added public property `$code_span_content_func`.

This adds an optional code span content function to use customized renderers for code span.

Just like the option `$code_block_content_func`, but now for code-span. Can be used to prevent the default `htmlspecialchars` or do some kind of other logic to convert the code to html.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -67,6 +67,9 @@ class Markdown implements MarkdownInterface {
 	
 	# Optional function for converting code block content to HTML
 	public $code_block_content_func = null;
+	
+	# Optional function for converting code span content to HTML.
+	public $code_span_content_func = null;
 
 	# Class attribute to toggle "enhanced ordered list" behaviour
 	# setting this to true will allow ordered lists to start from the index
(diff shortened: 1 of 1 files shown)


michelf:
Are you using this for syntax highlighting?

The code looks fine. I can probably merge this. I just want to know what you're using it for.

styxit:
Thank you, would help me a lot!

I have user submitted content. No html allowed, only markdown. I strip all html in advance, before giving it to the markdown parser. But when i use the code span, the content is stripped again. So in this case i want to disable the `htmlspecialchars` method for the code span.

**Original content, could be user submitted:**

```
Password: `123#aa>X`
```

**I strip special characters so it becomes:**

```
Password: `123#aa&gt;X`
```

**Desired output:**

```
<p>Password: <code>123#aa&gt;X<code></p>
```

**php-markdown output:**

```
<p>Password: <code>123#aa&amp;gt;X</code></p>
```

The `>` character in the password is cleaned by my own code to `&gt;` this code is cleaned by markdown parser code span to `&amp;gt;`

michelf:
But are you sure this is a safe way to sanitize user content? This is going to break some part of the markdown syntax that uses `<`, namely automatic links. It also won't block malicious things expressed in Markdown, such as `[some link](javascript:run_dangerous_script())`.

If all you want to do is to remove HTML tags, there's already an option for that called `no_markup`. It's not enough to protect you against a malicious script injection, but I think it does what you're trying to do.

The only safe way to sanitize potentially malicious content is to pass the output through a sanitizer, not the input.

styxit:
I know the `[some link](javascript:run_dangerous_script())` is still a threat. After running the content through the `transform` method i can do stuff like, remove links or something. This was only meant as an addition to specify the way the code span outputs the content it is given.

michelf:
I think I'll merge this because it makes sense for code highlighting. It also mirrors quite nicely what we already have for code blocks.

I'm still a bit worried that the way you're removing tags before parsing the content with Markdown could break parts of the syntax or leave security holes. But if you know what you're doing, that's your call to make.

styxit:
Thanks! We decided to only parse markdown for admin-users to prevent parsing malicious content.

PR #229, opened 2016-03-03
[#98] Changing to PHPDoc comments

robbieaverill:
Re-opens and completed pull request #100 

Addresses issue #98 to change to the PHPDoc style of code comments and documentation.

### DIFF
--- a/Michelf/Markdown.inc.php
+++ b/Michelf/Markdown.inc.php
@@ -1,10 +1,16 @@
 <?php
-
-# Use this file if you cannot use class autoloading. It will include all the
-# files needed for the Markdown parser.
-#
-# Take a look at the PSR-0-compatible class autoloading implementation
-# in the Readme.php file if you want a simple autoloader setup.
+/**
+ * Use this file if you cannot use class autoloading. It will include all the
+ * files needed for the Markdown parser.
+ *
+ * Take a look at the PSR-0-compatible class autoloading implementation
+ * in the Readme.php file if you want a simple autoloader setup.
+ *
+ * @package   php-markdown
+ * @author    Michael Fortin <michael.fortin@michaelf.com>
+ * @copyright 2004-2016 Michel Fortin <http://michelf.com/projects/php-markdown/>
+ * @copyright (Original Markdown) 2004-2006 John Gruber <http://daringfireball.net/projects/markdown/>
+ */
 
 require_once dirname(__FILE__) . '/MarkdownInterface.php';
 require_once dirname(__FILE__) . '/Markdown.php';
--- a/Michelf/Markdown.inc.php
+++ b/Michelf/Markdown.inc.php
@@ -7,7 +7,7 @@
  * in the Readme.php file if you want a simple autoloader setup.
  *
  * @package   php-markdown
- * @author    Michael Fortin <michael.fortin@michaelf.com>
+ * @author    Michel Fortin <michel.fortin@michelf.com>
  * @copyright 2004-2016 Michel Fortin <http://michelf.com/projects/php-markdown/>
  * @copyright (Original Markdown) 2004-2006 John Gruber <http://daringfireball.net/projects/markdown/>
  */
--- a/License.md
+++ b/License.md
@@ -1,5 +1,5 @@
 PHP Markdown Lib  
-Copyright (c) 2004-2015 Michel Fortin  
+Copyright (c) 2004-2016 Michel Fortin  
 <https://michelf.ca/>  
 All rights reserved.
 
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -1,105 +1,145 @@
 <?php
-#
-# Markdown  -  A text-to-HTML conversion tool for web writers
-#
-# PHP Markdown  
-# Copyright (c) 2004-2015 Michel Fortin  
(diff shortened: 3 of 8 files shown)


robbieaverill:
cc @itafroma - your input would be valuable.

Sorry it's a meaty code review!

michelf:
Indeed, can't review this on Github because the diff is too big. I'll take a look at a local copy later today or tomorrow.

Michelf/Markdown.inc.php
```diff
@@ -1,10 +1,16 @@
...
-# in the Readme.php file if you want a simple autoloader setup.
+/**
+ * Use this file if you cannot use class autoloading. It will include all the
+ * files needed for the Markdown parser.
+ *
+ * Take a look at the PSR-0-compatible class autoloading implementation
+ * in the Readme.php file if you want a simple autoloader setup.
+ *
+ * @package   php-markdown
+ * @author    Michael Fortin <michael.fortin@michaelf.com>
```
michelf: That should actually be `Michel Fortin <michel.fortin@michelf.com>`. There's no "a" in my name.

robbieaverill: Sorry! My mistake!

On Friday, 4 March 2016, Michel Fortin notifications@github.com wrote:

> In Michelf/Markdown.inc.php
> https://github.com/michelf/php-markdown/pull/229#discussion_r54891650:
> 
> > @@ -1,10 +1,16 @@
> > 
> > ##  <?php
> > 
> > -# Use this file if you cannot use class autoloading. It will include all the
> > -# files needed for the Markdown parser.
> > -#
> > -# Take a look at the PSR-0-compatible class autoloading implementation
> > -# in the Readme.php file if you want a simple autoloader setup.
> > +/**
> > - \* Use this file if you cannot use class autoloading. It will include all the
> > - \* files needed for the Markdown parser.
> > - *
> > - \* Take a look at the PSR-0-compatible class autoloading implementation
> > - \* in the Readme.php file if you want a simple autoloader setup.
> > - *
> > - \* @package   php-markdown
> > - \* @author    Michael Fortin <michael.fortin@michaelf.com <javascript:_e(%7B%7D,'cvml','michael.fortin@michaelf.com');>>
> 
> That should actually be Michel Fortin <michel.fortin@michelf.com
> <javascript:_e(%7B%7D,'cvml','michel.fortin@michelf.com');>>. There's no
> "a" in my name.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/michelf/php-markdown/pull/229/files#r54891650.

## 

_Robbie Averill_
LA PETITE MANOUCHE - NZ Gypsy Jazz Guitar Duo
Phone: +64 27 321 4758
contact@lapetitemanouche.com
www.lapetitemanouche.com

michelf:
Quick question about `@api`: shouldn't it apply to public configuration variables too? Or am I mistaken about the signification of this tag?

michelf:
This is actually bigger than I expected. It does look good overall. I was surprised that you added documentation for all the callback functions as they are mearly implementation details of the regex parsing. Those added comments and don't say much that isn't already obvious so I'm not sure if I should keep them, but for now they're in.

There are some comments that start with `/**` inside the function bodies. Is that intentional? Doesn't look harmful, so I'm merging anyway. There can always be followup pull requests for that.

Before merging, I corrected a syntax error you introduced on line 297 of MarkdownExtra.php. In short, you can't use `.` to concatenate two strings forming the initial value of a property, only a single string token is allowed for the initializer. We can't cut that line in two, sorry.

I also reverted some changes to Readme.php while merging. This file is intended to be sample code and I want to keep it short and uncluttered as much as possible.

robbieaverill:
Great. Thanks for looking over it so thoroughly.

I took the approach of documenting every function to be thorough. The
comments that were vague were because I wasn't sure exactly what the regex
was doing.

In regards to the API tag, I'm not sure about public properties. I think
best practice would suggest providing a public setter with API tag for them
and making them protected, but didn't want to change anything like that in
this pull request.

If nothing else I hope that this work will at least create a starting point
for future contributors to add to the documentation!

Cheers Michel

Robbie

On Saturday, 5 March 2016, Michel Fortin notifications@github.com wrote:

> This is actually bigger than I expected. It does look good overall. I was
> surprised that you added documentation for all the callback functions as
> they are mearly implementation details of the regex parsing. Those added
> comments and don't say much that isn't already obvious so I'm not sure if I
> should keep them, but for now they're in.
> 
> There are some comments that start with /*\* inside the function bodies.
> Is that intentional? Doesn't look harmful, so I'm merging anyway. There can
> always be followup pull requests for that.
> 
> Before merging, I corrected a syntax error you introduced on line 297 of
> MarkdownExtra.php. In short, you can't use . to concatenate two strings
> forming the initial value of a property, only a single string token is
> allowed for the initializer. We can't cut that line in two, sorry.
> 
> I also reverted some changes to Readme.php while merging. This file is
> intended to be sample code and I want to keep it short and uncluttered as
> much as possible.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/michelf/php-markdown/pull/229#issuecomment-192504551.

## 

_Robbie Averill_
LA PETITE MANOUCHE - NZ Gypsy Jazz Guitar Duo
Phone: +64 27 321 4758
contact@lapetitemanouche.com
www.lapetitemanouche.com

PR #231, opened 2016-05-12
Add "p" and "li" elements to $auto_close_tags_re

frostbitten:
https://www.w3.org/TR/html5/syntax.html#optional-tags

I never close these in my own coding and wouldn't expect to need to do it now.


frostbitten:
Actually, this isn't the appropriate line to change. In my tests I got an expected visual result, but that was because it was just creating empty paragraphs as spacers.

PR #232, opened 2016-05-19
Remove example file from autogenerated ZIP archives

mlocati:
This example file is not useful for production machines, so let's remove it from autogenerated ZIP archives.

Developers (the only ones that are interested in this) can still to a `git clone` (or simply take a look at the repo on GitHub :wink: ).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/Readme.php export-ignore


michelf:
The assumption here is that the zip archive is only, or mostly, used on production machines. I'm not really sure this is the case. Also, I wouldn't be surprised if plenty of production servers are built using git too.

Is the underlying problem you are trying to solve that Readme.php potentially lands in the PHP include path?

mlocati:
> The assumption here is that the zip archive is only, or mostly, used on production machines. I'm not really sure this is the case.

The approach of making clean distribution zip files is where most of the more important libraries are going (zend, symphony, ...).

> Also, I wouldn't be surprised if plenty of production servers are built using git too.

Maybe someone uses this approach, but most use composer.

> Is the underlying problem you are trying to solve that Readme.php potentially lands in the PHP include path?

No, it's just to have a clean set of libraries without any unneeded test/sample files (so that we don't have to manually remove them in production servers).
Developers that need sample files can go here on github.

mlocati:
Closing because of lack of interest from the repository owner

Issue #169, opened 2014-07-14
Automatic return-based linebreaks instead of “two spaces at end of line” linebreaks


solitud:
Hi, is there an option to turn on _Automatic return-based linebreaks instead of “two spaces at end of line” linebreaks_ like in the GitHub Markdown parser? If not are there any plans to implement it?

I know this is discussed controversially but I think it would be a huge benefit for vanilla users, and it would help to use Markdown syntax in Forum software etc.

standaniels:
Why is this controversially? Could you elaborate, @michelf?

I agree with @solitud, I think `return` character in plain text should create a `<br>` in HTML, because it breaks a line, doesn't it?

solitud:
@standaniels: I think it's not in the original specifications.
Anyway, if you dare to mess with the source you find it in _Markdown.php_ line 518:

```
protected function doHardBreaks($text) {
# Do hard breaks:
return preg_replace_callback('/ {2,}\n/', 
array($this, '_doHardBreaks_callback'), $text);
}
```

change to:

```
protected function doHardBreaks($text) {
# Do hard breaks:
return preg_replace_callback('/\n/',
array($this, '_doHardBreaks_callback'), $text);
}
```

michelf:
Because that's what the [Markdown Syntax](https://daringfireball.net/projects/markdown/syntax#p) document has to say about line breaks:

> The implication of the "one or more consecutive lines of text" rule is
> that Markdown supports "hard-wrapped" text paragraphs. This differs
> significantly from most other text-to-HTML formatters (including Movable
> Type's "Convert Line Breaks" option) which translate every line break
> character in a paragraph into a `<br />` tag.
> 
> When you _do_ want to insert a `<br />` break tag using Markdown, you
> end a line with two or more spaces, then type return.
> 
> Yes, this takes a tad more effort to create a `<br />`, but a simplistic
> "every line break is a `<br />`" rule wouldn't work for Markdown.
> Markdown's email-style [blockquoting](#blockquote) and multi-paragraph [list items](#list)
> work best -- and look better -- when you format them with hard breaks.

standaniels:
Ah, yes, other tags might break when `\n\r` == `<br>`, didn't think of that.

Is t a bug then, that when I want to start a list after a `<br>` it won't be created? For example:

```
Test<space><space>  
1. Option
2. Option
3. Option
```

Results in:

```
<p>Test<br>
1. Option
2. Option
3. Option</p>
```

[...]

PR #244, opened 2016-08-27
[#169] Adding hard wrap option

jcohlmeyer:
Adding hard wrap option to enable line breaks without two trailing spaces.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -69,6 +69,13 @@ class Markdown implements MarkdownInterface {
 	public $no_markup   = false;
 	public $no_entities = false;
 	
+
+	/**
+	 * Change to `true` to enable line breaks on \n without two trailling spaces
+	 * @var boolean
+	 */
+	public $hard_wrap = false;
+
 	/**
 	 * Predefined URLs and titles for reference links and images.
 	 * @var array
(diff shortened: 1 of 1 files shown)


michelf:
Being not sure what to do to test this, I ran `mdtest` against the parser with `hard_wrap` set to `true` and looked at the results. Obviously a lot of the tests are failing because the expected output does not match this behavior. But the interesting thing to look at is the output to determine whether it makes some sense.

For instance, it appears that this input:

```
*   Tab
    *   Tab
        *   Tab
```

produces a lot of extra `<br>` where it should not produce any. The situation is a bit similar for the handling of definition lists in MarkdownExtra. There's probably some extra newlines added for an intermediary step when parsing those lists that get converted to those extra `<br>`.

michelf:
I just want to add that while what I wrote above might seem a bit gloomy, nested lists and definition lists are the only two instances where I could see something was wrong. With those two exceptions, this new option seems to work as intended.

jcohlmeyer:
@michelf thanks for testing. Is there any way to easily determine the parent element from the doHardBreaks function so we can test against it to avoid these cases?

Any other ideas? I would love to help more but I am not sure where to go from here.

michelf:
There is unfortunately no tree structure or parent element you can look at. This is all nested string processing and it's a bit ugly to work with.

We need to figure out what is the input of `doHardBreak`, figure out why it gets that input, find a way to fix the input upstream or alternatively find a way to work around the issue in `doHardBreak`.

michelf:
Apparently I was wrong on one point when interpreting the results of parsing the test suite: definition lists work fine.

As for ordered and unordered lists, I think I have a way to fix the problem in the list parser. So I'll merge your change and push the necessary fix at the same time. It shouldn't take too long.

PR #245, opened 2016-08-28
Add code_tag_in_pre option

babeuloula:
With SyntaxHighlighter (Alex Gorbatchev), we don't need the `code` tag.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -93,6 +93,12 @@ class Markdown implements MarkdownInterface {
 	 * @var callable
 	 */
 	public $code_block_content_func = null;
+	
+	/**
+	 * If we need the code tag (SyntaxHighlighter)
+	 * @var boolean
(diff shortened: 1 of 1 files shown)


michelf:
Are you saying that SyntaxHighlighter gets confused by the `<code>` tag? Assuming it is so, I'm honestly not sure the fix should be applied on PHP Markdown's side: SyntaxHighlighter should be fixed so it works with `<code>` too.

babeuloula:
I have this with the plugin : 

image

michelf:
So it does get confused. Any idea whether it can be fixed in SyntaxHighlighter? I think that'd be the better option. I'd rather not add configuration variables for each little output variant people need. It's a mess to properly maintain and it makes things more complicated for users of the library.

babeuloula:
Ok as you want. I'll leave a message on his github.

PR #246, opened 2016-08-28
Add code_tag_in_pre option

babeuloula:
With SyntaxHighlighter (Alex Gorbatchev), we don't need the `code` tag.


mclassic:
Duplicate of #245

PR #254, opened 2016-10-30
Readme: highlight syntax for PHP

TomasVotruba:


michelf:
Not a bad idea in itself. But I think i'll pass for two reasons:
1. That Readme.md file is used as a demo for using the `Markdown` parser in Readme.php, and only `MarkdownExtra` can parse fenced code blocks like those.
2. When reading it as a text file, the text and code is more readable using indented code blocks the way it is currently.

TomasVotruba:
1. README should not be limited to repository source code abilities.
   Better place for that is a test. Can do.
2. How often do you look to README source over README itself? Me only for PRs.
   Code (text) is read 10x times more then written. This PR makes it smooth to read and scan as any code is highlighted. This is improvement for users reading the README file and it can improve their experience with this package.

michelf:
The code is only highlighted when viewing the repository over Github. If you have have a copy on disk and open the file in an editor, you can only see the unprocessed textual form, the "source". One of Markdown's original goal is for the text be natural to read in its unprocessed form, and true to that making the unprocessed Readme file easy to read is a goal to illustrate what Markdown is.

> [The idea is that a Markdown-formatted document should be publishable as-is, as plain text, without looking like it’s been marked up with tags or formatting instructions.](https://daringfireball.net/projects/markdown/) -- John Gruber

There are multiple ways to use this parser, not all of them involve Github, and not all of them involves using the extensions to Markdown like fenced code blocks where you can specify the syntax to highlight. The Readme reflects that both in prose and in form.

If this was any other project, I'd merge without hesitation. But this is a Markdown parser and out of principle the Readme should follow the philosophy of the project and the parser ought to be able to parse its own Readme file. As long as I do no phase out the `Markdown` parser without extensions, I'll also be willing to make the small trade-off of not having syntax highlighting when viewed through Github.

TomasVotruba:
I don't want to argue about Markdown purity, nor support in your favorite IDE/editor for this file. This is not my goal here.

**I only use available tools provided by Github to make the best reading README experience.** To make readers **feel better** as they would while reading code almost anywhere else.

michelf:
I know you only want good. Note that I'm not criticizing the fact that you proposed this: it's a valid proposal that makes sense. It just goes in a different direction to the goals I had when writing that file in the first place, goals that are quite specific to this project. The proposal you make here asks me to renounce to those goals, and your argumentation is that they are less important than the syntax highlighting when looking at the Readme file on Github. I'm still unconvinced.

It's possible those goals makes no sense to you, in which case it's normal we differ in opinion. Beyond explaining the rationale as I did earlier, I don't know what I can do to reconcile our points of view. The only question that comes to mind is why are you so disturbed by the lack of syntax highlighting? Are those two-line snippets really harder to read for you without the code highlighting? I guess it's at least an annoyance otherwise we wouldn't be arguing about this here. Or does it go beyond that?

TomasVotruba:
> Are those two-line snippets really harder to read for you without the code highlighting?

Yes, that's my point. It's the same as opening this website code in notepad and in IDE/Editor.
You have written the code, so it's easy to read for you. But I read it for the first time. I see README as entry point for end users, so IMO should be as helpful as possible. Not less.

Compare yourself:

```
$this->collectByType(ConsoleApplication::class, Command::class, 'add');
if (!$m = Strings::match($file->getFilename(), '~^(?P<domain>.*?)\.(?P<locale>[^\.]+)\.(?P<format>^\.]+)$~')) { 
    continue;
}
$this->collectByType(RouteDecorator::class, RouteInterface::class, 'addRoute');
```

to

``` php
$this->collectByType(ConsoleApplication::class, Command::class, 'add');
if (!$m = Strings::match($file->getFilename(), '~^(?P<domain>.*?)\.(?P<locale>[^\.]+)\.(?P<format>^\.]+)$~')) { 
    continue;
}
$this->collectByType(RouteDecorator::class, RouteInterface::class, 'addRoute');
```

---

**What is you real goal**, if we step away from README file itself? I don't understand yet and it seems you try to bend README to some other purpose, probably testing (my bare guess).

michelf:
I'm not sure what there is to discuss outside of the Readme file. Yes, I'm cramming two or three different usages into that Readme file: online reading on Github, offline reading in a text editor, and I use the file as the Markdown source in Readme.php as example of how to use the parser. I don't think it's stupid to overload those three usages.

The only drawback of this is the lack of syntax highlighting on Github, which I see as a minor annoyance I'm willing to accept given the fact that it makes the file better suited for its other two purposes. 

A possible solution would be to have a separate Readme.text for the two other purposes that is formatted in plain Markdown without extension. That way you'd get syntax highlighting on Github and the text version can be used for the other two purposes. But, I don't thrilled at the idea of maintaining two Readme files.

At this point it's probably better to acknowledge we disagree on this issue and move on. This discussion does not seem to be advancing anything.

TomasVotruba:
I recommend using Travis for tests. It brings validation value to the code that is missing now.

michelf:
This has nothing to do with testing. 😞

(And if for some reason you are looking for tests, there're in [MDTest](https://github.com/michelf/mdtest).)

TomasVotruba:
Well, you test something manually and visually.

Why moving test to another repository? This is very confusing for me.

Standard is being in same repository in `/Tests` or rather `/tests` directory.

How do you verify that merge didn't break anything?

michelf:
I don't mind discussing the test suite, but don't you think it deserves its own issue?

TomasVotruba:
Ok. Feel free to create one.

michelf:
See #256

PR #255, opened 2016-10-30
composer: drop deprecated alias

TomasVotruba:
I recommend dropping alias, since it has not enough attention to keep them up to date. Usual case for cases most I've seen.
Now it's version 1.8-dev so it's been useless for a while now.


michelf:
Makes sense.

TomasVotruba:
Thanks

PR #265, opened 2017-04-13
Be more specific about PHP versions

Nyholm:
We do not support PHP6 nor PHP8. We should state that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
 		}
 	],
 	"require": {
-		"php": ">=5.3.0"
+		"php": "^5.3 || ^7.0"
 	},
 	"autoload": {
  		"psr-0": { "Michelf": "" }


michelf:
Why are you proposing this?

As far as I know, there has been no release of PHP 6 and PHP 8 isn't a thing yet. Why should we care about those? Is there a reason to expect PHP Markdown to not work with PHP 8?

vinkla:
> Is there a reason to expect PHP Markdown to not work with PHP 8?

The simple answer is that we don't know yet. If anything will break in PHP version 8 this package (at its current state) will still be downloadable.

Nyholm:
Why? Because we are not testing for PHP6 or PHP8. How can we promise we still are compatible with those? 

> Why should we care about those?

We should not, that is why I made this PR. ;)

This is just a semantic change, It will have no real effect until PHP8 is actually released. Im okey with closing it if you do not like it.

michelf:
To tell you the truth, I think I never tested it with PHP 7.x. Maybe then I should remove PHP 7 until I take the time to test it properly... but surely plenty of users are using it without problem with PHP 7 by now. I can't know _for sure_ I have users on PHP 7 as I got no PHP 7-specific bug report, but I'd say it's quite likely, and I'd be a disservice to them to remove PHP 7 from that list.

How I'm going to handle the PHP 8 transition when it comes? It'll probably follow the same pattern of no proactive testing and it'll probably be fine. If not I'll get bug reports and fix them quickly.

My opinion is that it could make sense to adopt a pattern of whitelisting the versions known to work, but it'd have to be accompanied with a more elaborate testing and release management process.

Nyholm:
I claim that this is not fixed.

PR #267, opened 2017-04-17
Refactor create_function into anonymous function

langemeijer:
This is good for readability. The values in the regexp had to be escaped twice.

Anonymous functions is available since PHP 5.3, which is the minimum version for php-markdown anyway, so using this will not increase the version number dependency.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -1872,7 +1872,7 @@ class Markdown implements MarkdownInterface {
 
 		$this->utf8_strlen = function($text) {
 			return preg_match_all('/[\x00-\xBF]|[\xC0-\xFF][\x80-\xBF]*/', $text, $m);
-		}
+		};
 	}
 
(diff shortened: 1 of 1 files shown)


Michelf/Markdown.php
```diff
@@ -1870,9 +1870,9 @@ protected function _initDetab() {
 			return;
 		}
 
-		$this->utf8_strlen = create_function('$text', 'return preg_match_all(
-			"/[\\\\x00-\\\\xBF]|[\\\\xC0-\\\\xFF][\\\\x80-\\\\xBF]*/", 
-			$text, $m);');
+		$this->utf8_strlen = function($text) {
+			return preg_match_all('/[\x00-\xBF]|[\xC0-\xFF][\x80-\xBF]*/', $text, $m);
+		}
```
michelf: This does not parse: it's missing a semicolon.

PR #269, opened 2017-04-22
Clean up trailing white space across all files

jdufresne:


michelf:
That's probably a good idea for the PHP files. 

It's a bit trickier in Markdown files however. While there are many unneeded trailing spaces, some of them are actually a double-space-before-a-newline which in some contextes is the Markdown syntax for a line break. At a minimum, those should be preserved.

jdufresne:
I've restored intentional line breaks in the MD files.

What do you think about replacing the trailing spaces with a `<br>` tag? Many editors routinely clean up trailing white space. Using trailing white space as syntax breaks this behavior. Also it is difficult to see as white space is normally not visible. Additionally, the license file could easily be a plain text file as it has no real formatting. Most license files are. If not, I think this is merge ready.

michelf:
Given this is a project specifically about Markdown, it's important to set an example by making the Markdown-formatted files in the project follow its philosophy as much as possible. Those files should be as pleasant to read in plain text as they are in rendered HTML. That's how this double-space syntax came by in the first place: it's invisible and non-distracting to the reader in plain text.

I'll admit I myself messed up those trailing whitespaces a few times in Markdown documents because of editors that trimmed them. I think for this particular project it's worth maintaining those whitespace-linebreak as they are.

Issue #202, opened 2015-05-06
Documentation


ghost:
**Readme.md**: The _Usage without an autoloader_ section might be more instructive as follows:
### Usage without an autoloader

If you cannot use class autoloading, you can still use `include` or `require` 
to access the parser. To load the `\Michelf\Markdown` parser, do it this way:

```
require_once 'Michelf/Markdown.inc.php';
print \Michelf\MarkdownExtra::defaultTransform("*Test*");
```

Or, if you need the `\Michelf\MarkdownExtra` parser:

```
require_once 'Michelf/Markdown.inc.php';
require_once 'Michelf/MarkdownExtra.inc.php';   
print \Michelf\MarkdownExtra::defaultTransform("*Test*");
```

While the plain `.php` files depend on autoloading to work correctly, using the
`.inc.php` files instead will eagerly load the dependencies that would be 
loaded on demand if you were using autoloading.

ghost:
Typical user error. Please disregard.

Issue #135, opened 2013-11-28
PHP MarkdownExtra as single file


markseuffert:
At the moment `\Michelf\MarkdownExtra` is implemented as a subclass of `\Michelf\_MarkdownExtra_TmpImpl`. Are there any plans to change this? When using PHP Markdown Extra it would be nice to be able to use a single PHP file, as in [previous release](http://michelf.ca/projects/php-markdown/classic/). Thanks!

michelf:
Actually yes, there is a plan to move the actual implementation of MarkdownExtra to the MarkdownExtra.php file. I've only temporarily refrained from doing so because I want git merges to continue to work between the extra and lib branches. After the cutoff date for the classic parser and the extra branch (February 1, 2014) I'll move everything where they logically belong.

Do you mean you want to `include` only one file? I'm aware that not everyone use autoloading and I plan to have a "master include" file that'd include the others. With the changes I just made today in the lib branch you have to include both MarkdownInterface.php, Markdown.php, and MarkdownExtra.php, and in the right order, if you wish to use the MarkdownExtra parser. That's not pretty. The only thing delaying that is I wonder what should be the name of that file and whether it should be put in the Michelf directory or at the root. If you've seen this "master include" pattern elsewhere, or something else doing the same job, please let me know.

markseuffert:
Yes, I would like to include only one file. It was quite handy in the past. I didn't know that the plan is to have multiple files in the future, so I probably wait until February 2014 and then see what's a good way to switch to latest MarkdownExtra version.

markseuffert:
Switching to 1.4.0 was very simple. Included `Markdown.php` with minor modifications, renamed class and removed interface. If Markdown Extra class would not be abstract and interface would be in same file, it could be used without any modification.

michelf:
But have you tried including the ".inc.php" file?

https://github.com/michelf/php-markdown#usage-without-an-autoloader

markseuffert:
Then I would have to add 4 files to my project. `MarkdownExtra.inc.php` plus the 3 referenced in it.

michelf:
Ah, I see. There's no escape to the fact that the "lib" distribution is a folder and not a single file, even though you can now `require_once` only one file.

The abstract class in the Markdown.php file is a temporary hack to be able to merge the changes between the extra and lib branches. It's abstract because you shouldn't use it.

You really can't add a folder with 6 files in it to your project instead of adding a file?

jaswrks:
[...]

PR #272, opened 2017-04-23
Use PSR-4 for autoloading

jdufresne:
This change brings this library's autoloading up to modern standards in the larger PHP community and uses methods recommended by the PHP dependency manager, composer.

From PSR-0 http://www.php-fig.org/psr/psr-0/

> As of 2014-10-21 PSR-0 has been marked as deprecated. PSR-4 is now recommended as an alternative.

PSR-4: http://www.php-fig.org/psr/psr-4/

From composer https://getcomposer.org/doc/04-schema.md#autoload

> PSR-4 is the recommended way since it offers greater ease of use (no need to regenerate the autoloader when you add classes).

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -43,17 +43,28 @@ Before PHP 5.3.7, pcre.backtrack_limit defaults to 100 000, which is too small
 in many situations. You might need to set it to higher values. Later PHP
 releases defaults to 1 000 000, which is usually fine.
 
+Installation
+------------
+
+Install this library with composer:
+
+	$ composer require michelf/php-markdown
 
 Usage
 -----
 
-This library package is meant to be used with class autoloading. For autoloading
-to work, your project needs have setup a PSR-0-compatible autoloader. See the
-included Readme.php file for a minimal autoloader setup. (If you cannot use
-autoloading, see below.)
+This library package is meant to be used with
+composer's
+[autoloading](https://getcomposer.org/doc/01-basic-usage.md#autoloading). Simply
+include composer's generated `vendor/autoload.php`:
 
-With class autoloading in place, putting the 'Michelf' folder in your
-include path should be enough for this to work:
+	require 'vendor/autoload.php';
+
+Without composer, for autoloading to work, your project needs a
+PSR-4-compatible autoloader. See the included Readme.php file for a minimal
+autoloader setup. (If you cannot use autoloading, see below.)
+
+With class autoloading in place:
 
 	use Michelf\Markdown;
 	$my_html = Markdown::defaultTransform($my_text);
--- a/Readme.php
+++ b/Readme.php
@@ -4,9 +4,9 @@
 // through the Markdown filter. You can adapt this sample code in any way
 // you like.
 
-// Install PSR-0-compatible class autoloader
+// Install PSR-4-compatible class autoloader
 spl_autoload_register(function($class){
-	require preg_replace('{\\\\|_(?!.*\\\\)}', DIRECTORY_SEPARATOR, ltrim($class, '\\')).'.php';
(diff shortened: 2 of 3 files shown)


michelf:
I have no objections to improving the documentation to explain a bit more how to set it up correctly with Composer. But most of these changes are unacceptable.

1. Many people do use the library without autoloading. That's why the ".inc.php" files were added in the first place. Removing them is a breaking change with negligible benefits.

2. The same could be said of moving the files to a "src" directory. It would break the package for any user not depending on Composer. That would make the package incompatible with previous versions.

3. I don't wan't the Readme.php file to necessitate the package to be installed in a particular way. For instance, it should work out of the box when unpacking the zip archive and running the script. That's why I wrote that minimal autoloading function.

While PHP Markdown must work with Composer, it is not my intention to recommend a particular packaging system. Not everyone has the luxury of working with projects that follow modern standards, so I choose to be accommodating.

jdufresne:
I've reworked the PR.

* Restored source directory name to `Michelf`
* Restored `*.inc.php` files
* Restored Readme.md sections about using this library without autoloading.
* `Readme.php` no long uses composer's `autoload.php`

The PR now uses PSR-4, but doesn't pull in the other community conventions of the original version.

michelf:
The last time an autoloading file was added to this project is in 2013 and I don't foresee adding more soon. The gains we could get from PSR-4 don't really apply here, but there's nothing really holding us back either. Composer added support for PSR-4 in [January 2014], 3 years ago. I think that's good enough. Anyone using a version that old by now is likely running into problems with all the PSR-4 packages out there. So I'll accept the change for PSR-4.

 [January 2014]: https://github.com/composer/composer/blob/master/CHANGELOG.md#100-alpha8---2014-01-06

I like that the instructions are clearer now for Composer. I'll change it a bit after merging however because I think combining this new text with the previous one makes it a bit more verbose than necessary.

jdufresne:
> Many people do use the library without autoloading.
> ...
> Not everyone has the luxury of working with projects that follow modern standards,

I find this claim to be surprising. Composer is used by every PHP project and library I have come in contact with for several years now. Composer has clearly emerged as the community's preferred package manager and continues to be well maintained. A project that avoids Composer would be putting themselves at a serious disadvantage. I can't recall any other library I've used that didn't simply rely on Composer's `autoload.php`. So while such a project would work well with php-markdown, it couldn't integrate easily with virtually any other PHP library.

A project that truly needs to stay away modern PHP conventions could continue to use an old version of php-celery.

Thanks for considering. Cheers.

michelf:
I [sometime get thanks][thanks] for those `.inc.php` files. And I know that some people are still using it (and sometime [fail to follow the instructions][fail]). Unfortunately, I don't get much feedback and I can't compile statistics about how people are using the library or how Composer usage changed over the years.

 [thanks]: https://github.com/michelf/php-markdown/issues/135#issuecomment-30839650
 [fail]: https://github.com/michelf/php-markdown/issues/202

I'm pretty sure there are many projects and websites out there that aren't open source, are only sporadically maintained, and still don't use Composer. Sometime people want old things to continue to work without requiring constant refactoring to follow the current recommended standards. Forcing them to use Composer to use a newer PHP Markdown is a breaking change and would require a 2.0.0 version if I was to respect semantic versioning. I'm not going to bump the major version and drop backward compatibility just for the benefit of shuffling the files to make it looks more modern. Moreover cost of keeping that compatibility is negligible, so why get rid of it?

I think we can keep that idea around for when we do have a good reason to make a 2.0 however.

michelf:
I don't necessarily want the argumentation to be about me, but I should probably also mention my bias: I have a few decade-old small PHP scripts on my mostly-static website that use PHP Markdown and aren't relying on Composer. I'm definitely one of those benefiting from these backward compatibility affordances.

PR #277, opened 2017-06-13
Add an option to enforce the space when converting atx-style headers

### LINKED ISSUE #275: Don't transform to a header if no space after #
Is it possible to disable the title configuration if there is no space between the # and the first character of the title? In some cases people are declaring hashtags this way.

See https://github.com/jupyter/notebook/issues/474, http://www.cirosantilli.com/markdown-style-guide/#option-header-atx and https://daringfireball.net/projects/markdown/basics

edhelas:
Fix #275

### DIFF
--- a/Michelf/MarkdownExtra.php
+++ b/Michelf/MarkdownExtra.php
@@ -77,7 +77,7 @@ class MarkdownExtra extends \Michelf\Markdown {
 	 * Only convert atx-style headers if there's a space between the header and #
 	 * @var boolean
 	 */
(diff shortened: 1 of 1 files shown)


michelf:
Looks good. Ruminating this a bit, I think the name `enforce_space_atx_style_headers` is not specific enough. I'd prefer if it was called `hashtag_protection` (or something similar) so it reflects the purpose instead of describing the effect. The documentation comment describes the effect well enough.

edhelas:
Here we go :)

PR #284, opened 2017-09-29
Trim FencedCodeBlocks on MarkdownExtra.php

yidas:
Fix `<pre><code>` code block string with trim, make it standard such as GitHub markdown.


michelf:
Can you explain more clearly what this is trying to achieve? I see you are trimming the whitespace, but I don't understand why.

yidas:
For a markdown example code:

```
~~~
Code block
~~~
```

The current parser will parse to HTML as:

```
<pre><code>Code block
</code></pre>
```
This will leave a  newline in the end, which could make syntax highlighter has one more blank line in the last line.

After `trim()` the Code block, the result will be:

```
<pre><code>Code block</code></pre>
```

Which is same as GitHub markdown parse rule.

Thank you for your quickly reply.

michelf:
Are you suggesting this because there is a syntax highlighter out there that has problem with this?

Also, I haven't really tested it, but it seems to me that using `trim` would also remove intentional blank lines before and after the code, as well as indentation for the first line. All you need is to remove the last `\n`, not all the surrounding whitespace.

yidas:
Thanks for the recommendation.

I changed to `rtrim` with newline concern only.

Issue #256, opened 2016-10-31
Better Test Suite


michelf:
Currently, I use [MDTest](https://github.com/michelf/mdtest) to test PHP Markdown after each change. It's done manually with two commands:

    ./mdtest.php -f \\Michelf\\Markdown::defaultTransform -t "Markdown" -t "PHP Markdown"
    ./mdtest.php -f \\Michelf\\MarkdownExtra::defaultTransform -t "Markdown" -t "PHP Markdown" -t "PHP Markdown Extra"

The first line runs the Markdown and PHP Markdown test suites against the `Markdown` parser, while the seconds tests the same test suites, plus the PHP Markdown Extra suite, against the `MarkdownExtra` parser. (Note that php-markdown is a submodule inside MDTest, that's why it can find the parser classes.)

The output of MDTest tells me whether there was any error, comparing a normalized form of the output with the expected output. It can display a diff on-demand between the output and what was expected (with `-d`). It displays the time it takes to run each tests... which is ridiculously small these days because our CPUs are so fast, but is still useful to asses there is no significant performance regression. I also have a private test suite that includes some lengthy text files that works better to test speed regression that I generally add to the command above, but unfortunately those documents are not public.

I run MDTest as a validation step before pushing anything to the repository. Everything should pass, with one exception: the test for emphasis in the PHP Markdown test suite is incompatible with `MarkdownExtra` and the expectation is that it fails when testing `MarkdownExtra` (I know, it's clumsy). Then I assess that the speed is not abnormally slow. Once everything is good, I push to Github.

Now, this setup works well for me. But I think it's a put-off for contributions. There's also a few things that currently go untested and that could be improved: basically any non-default configuration of the parser. I wish I had a solution for this, but testing everything here quickly becomes a combinatory explosion.

So I'm open to suggestions. If you have a question about why something is done in a certain way, please ask. I don't have much time to put for improvements here, so most likely if there are changes to be made the implementation will have to be made by someone else.

*Some background on MDTest: the way it works is relic of a time I hoped to make a test suite that other implementations could use. If widely used, it could have helped reduce the fragmentation of Markdown in various parser-dependent syntaxes each with different quirks. That's why it's a command line tool capable of testing can test pretty much any other command line tool that emits some Markdown output. But I'm not sure there's much value in this setup anymore.*

michaelbutler:
OK, I will add to my ToDo list to explore the possibilities here. Thanks!

michaelbutler:
[...]

PR #291, opened 2018-01-27
Add MDTests and PHPUnit 5.7 to Run Them

michaelbutler:
For this PR, I've copied the test resources (input `.text` and their corresponding `.html`/`.xhtml` files) and added a unit test suite that can be run via PHPUnit. I've run `composer require --dev "phpunit/phpunit:5.7"` to get PHPUnit as a dev dependency. 

However at this point I'd like to ask about the tests in MDTest. Specifically, how up to date are they? Only about `50%` pass (meaning match the exact expected output). However, a lot of the failures are simply due to whitespace differences -- sometimes trailing new lines, and sometimes new line characters within the output. I guess this is what is meant by "normalized" or not. But why would it vary? Can the expectations simply be updated to match what the "true" output should be? Or does the markdown parser have options that can be tweaked to affect the output. 

Besides the whitespace differences, the remainder of the failures seem to be due to HTML encoding which I don't know why is happening. For example, given this input (`markdown/Backslash escapes.text`):

```
These should all get escaped:

Backslash: \\

Backtick: \`

Asterisk: \*

Underscore: \_
```

We expect to get `markdown/Backslash escapes.xhtml`:

```
<p>These should all get escaped:</p>

<p>Backslash: \</p>

<p>Backtick: `</p>

<p>Asterisk: *</p>

<p>Underscore: _</p>
```

What is actually produced by `Markdown::defaultTransform`:

```
<p>These should all get escaped:</p>

<p>Backslash: &#92;</p>

<p>Backtick: &#96;</p>

<p>Asterisk: &#42;</p>

<p>Underscore: &#95;</p>
```

Is it implied that the entities need to be decoded first before confirming the output is correct?

# Running the tests

The tests can be run by command line. After running `composer install` to download PHPUnit, run `vendor/bin/phpunit`.

If there's anything else you'd like to see or change, please let me know!


michelf:
When I run MDTest, all the test pass fine (except for the emphasis test which behaves differently with Extra). You need to use normalization to get that result though. The normalization process will adjust whitespace, standardize the order of attributes, and resolve character entities... perhaps it does other things I forgotten about too, it's been a long time.

There are situations where we could (and probably should) be more strict and require an exact match with the expected output. There are situations where you just want a general sanity check that a change in the parser isn't breaking existing documents, normalization is great because you don't have to inspect every little detail to check if the output is correct. And I believe there are some cases where the output will have some differences between Markdown and MarkdownExtra because of internal implementation details, differences that normalization would erase allowing us to reuse the same test for both.

I guess the best course of action long term would be to be to allow each test to decide whether it should match with normalization or without. But as a starting point, I would suggest using normalization everywhere as this is how I use MDTest at the moment.

_For future reference, this is related to #256_

michaelbutler:
Thanks a lot for the feedback! That clears up a lot. I'll add in the normalization for now and read more closely the Issue #256 .

michaelbutler:
Ok, after adding the same (x)html normalize comparison logic to the unit tests, all 60 comparison tests pass. You can confirm by running `vendor/bin/phpunit`, and with xdebug installed code coverage:
`mkdir build && php -dmemory_limit=512M vendor/bin/phpunit --coverage-html=./build/`

It's sitting at 97.64% which is quite good. Using PHP 7.0 by the way.

michaelbutler:
screenshot from 2018-01-27 21-01-28

michelf:
Looks quite good. I'll give it a try.

Did you think about running the tests found in "markdown.mdtest" and "php-markdown.mdtest" against `MarkdownExtra` too? They should all pass with both `Markdown` and `MarkdownExtra`.

There's one exception to the rule above: the "Emphasis" test in "php-markdown.mdtest". Because `MarkdownExtra` treats underscore emphasis differently, that test cannot and should not pass with `MarkdownExtra`. Perhaps that test should be moved to a separate test suite ("php-markdown-regular.mdtest") so it can be tested separately only against the `Markdown` parser.

michaelbutler:
> Did you think about running the tests found in "markdown.mdtest" and "php-markdown.mdtest" against MarkdownExtra too? They should all pass with both Markdown and MarkdownExtra.

Sounds good, will make some additional adjustments over the next 2 days.

michaelbutler:
@michelf apologies for the delay. I've added the additional runs you recommended: `markdown.mdtest` against `MarkdownExtra`, and `php-markdown.mdtest` against `MarkdownExtra` with the exception of `Emphasis` test, which I just excluded in the data provider for that test.

michelf:
Nice. Thank you. I'll give it a try and merge soon.

michaelbutler:
Cool. Once merged I can start working on a second PR to enable either Travis-CI or Circle-CI integration, which will auto run unit tests on new pull requests, calculate code coverage, and etc. Do you have a preference for any service? (They're all free to use for open source projects)

michelf:
No preference really. The CI I know is the one that comes with Xcode which I use for Mac and iOS projects. I know Travis and Circle by name, but never used them.

Issue #286, opened 2017-12-21
Placing footnotes where you want vs end of document


stefanocpp:
Hello,

I have a blog I made with PicoCMS. I write content with Markdown extra and often use footnotes. I'm looking for a way to place footnotes at the end of each article instead of having them at the end of the whole document, say having a string that tells where to put footnotes used to that point. In fact, having them placed in HTML where I place them in Markdown would be perfect.

thanks

michelf:
What a CMS usually do is parse each article separately using Markdown, then assemble the HTML output of each in one big page. I'm not sure how Markdown is integrated in PicoCMS, but what you describe sounds like it could cause problems if two articles share the same name for footnote or link references.

stefanocpp:
Hi, thanks for replying. What I have is a CMS that converts each page in a separate html page so there shouldn't be a problem of collision because every page would have its own set of footnotes being originated by just one source markdown document.
The problem I have now is that if I have separate articles in one big document, each one of them with its own set of footnotes, they all get placed at the end of the whole document even though I separate them with <hr> (***) or place the articles' footnotes at the end of each one of them.

thanks again!

PR #294, opened 2018-03-01
Allows footnotes to be placed elsewhere.

jkrrv:
I have a project where I need to be able to place the footnotes differently.  (Someone else suggested this in #286.)  I really do need it, so I figured I'd develop it and contribute the feature back. 

This change takes the existing code that generates the footnote `ol` and moves it to a separate, public function: `getFootnotes()`.  It also adds a configuration variable, `append_footnotes` which, when set to `false`, prevents the default behavior of appending the footnote list in a div at the end of the body. 

I'd be happy to adjust the nomenclature; there doesn't seem to be a precedent in the library for this sort of function.

Test results with this change in place are the same as test results without the change.

### DIFF
--- a/Michelf/MarkdownExtra.php
+++ b/Michelf/MarkdownExtra.php
@@ -81,10 +81,18 @@ class MarkdownExtra extends \Michelf\Markdown {
 
 	/**
 	 * Determines whether footnotes should be appended to the end of the document.
-	 * Footnotes can also be retrieved via the getFootnotes() function.
+	 * If true, footnote html can be retrieved from $this->footnotes_assembled.
 	 * @var boolean
 	 */
-	public $append_footnotes = true;
+	public $omit_footnotes = false;
+
+
+	/**
+	 * After parsing, the HTML for the list of footnotes appears here.
+	 * This is available only if $omit_footnotes == true.
+	 * @var null|string
+	 */
+	public $footnotes_assembled = null;
 
 	/**
 	 * Parser implementation
@@ -128,7 +136,6 @@ class MarkdownExtra extends \Michelf\Markdown {
 	 */
 	protected $footnotes = array();
 	protected $footnotes_ordered = array();
-	protected $footnotes_assembled = null;
 	protected $footnotes_ref_count = array();
 	protected $footnotes_numbers = array();
 	protected $abbr_desciptions = array();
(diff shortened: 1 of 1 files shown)


michelf:
This seems like a good idea. I wonder how you're placing your footnotes though. Also, I don't think this really fits with what was requested in #286.

You probably should reset `footnotes_assembled` in `setup()` so that if you reuse the parser to parse some other text you don't get the previous run's footnotes.

Also, did you test whether `getFootnotes()` works correctly when called after the parser has done its parsing? `footnotes_ordered` gets cleaned in `teardown()`, so I'd expect the list of footnotes to always be empty once the parsing is done and `append_footnotes` is `false`.

I think instead of `append_footnotes` I'd call the configuration variable `omit_footnotes` (with `false`=included, `true`=omitted). This would better highlight the behavior relative to the default.

And perhaps we should call the function `getOmittedFootnotes()` and say it returns `nil` when `omit_footnotes` is `false` and parsing is done. This would allow us to clear the parser's variables in `teardown()` and not retain more memory than necessary, unless `omit_footnotes` is `true`. What do you think?

jkrrv:
That all seems reasonable, and is all compatible with my use case.  I don't particularly care what the variables/functions are named, but I'd much rather change it now than have breaking changes later.  My only concern is that some of the particulars (e.g. clearing the variable in teardown if `!omit_footnotes`) aren't particularly obvious, and thus especially need appropriate attention in the documentation.  

My latest push (8bbbbf0) reflects your preferences, with one difference: with `footnotes_assembled` getting cleared in `teardown`, it has the same value as what you've requested for `getOmittedFootnotes()`.  Since it's overwritten immediately before any reads, I'm not concerned about users causing weird errors by setting that variable elsewhere. 

You asked how I'm placing my footnotes.  I'm working on a viewer for long documents (>100k words) which are divided into many relatively short chapters.  I want the user to be able to scroll past the end of a chapter and have the next chapter load via XHR.  But, I want the footnotes to appear be in a separate block on the page, with the contents of that block changing to the current chapter.  This is just easier to do if I can separate out the footnotes. 

Or, the more generic use case: sometimes you want the body and footnotes separated by something other than an `<hr />`.

michelf:
Interesting use case.

This looks all right. I'll give it a try soon and it should be ready to merge.

michelf:
I tested and I noticed something for which I'm not sure there's a great the solution. We're loosing the `<div>` that encloses the footnotes when looking at `footnotes_assembled` with this. I normally wouldn't mind this at all. But this means we're also loosing the `role="doc-endnotes"` attribute on that `<div>`.

I don't trust most users of the library will think of adding back the `role` attribute at the right place in their design to keep the accessibility intact. I don't think there's anything we can do in code to help that though. So I simply added some documentation to this effect in dfaa413.

For reference: #287

PR #296, opened 2018-03-19
Remove PHP 5.4 and 5.5 from auto testing and update README

michaelbutler:
I've removed PHP 5.4 and PHP 5.5 from the auto testing because it's too much hassle to get both old and new PHPUnit working in the same repository. The newer PHPUnit has to be used to test against PHP 7.0 and 7.1, so it's recommended that be the priority.

Also updated the README so the travis-ci build status is shown at the top.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ cache:
 
 # Run tests against all these PHP versions
 php:
-  - 5.4
-  - 5.5
   - 5.6
   - 7.0
   - 7.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ cache:
 
 # Run tests against all these PHP versions
 php:
+  - 5.4
+  - 5.5
   - 5.6
   - 7.0
   - 7.1
--- a/Readme.md
+++ b/Readme.md
@@ -1,14 +1,12 @@
 PHP Markdown
 ============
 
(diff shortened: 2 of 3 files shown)


michaelbutler:
One can see that Travis-CI has passed this build :heavy_check_mark:

michelf:
I was wondering whether it'd be reasonably achievable to fix this instead of disabling the tests. If we need two different versions of PHPUnit, then I agree it's too much hassle.

> Also updated the README so the travis-ci build status is shown at the top.

I've always disliked the aesthetics of those badges. Can we go without?

Also, I think your editor inadvertently removed all the trailing double spaces from the Readme. Those spaces are significant — they create line breaks — and so they should stay there.

> One can see that Travis-CI has passed this build ✔️

Yes! I was looking forward to your pull request just for that. 💯 Great work.

michaelbutler:
Understand your concern about the badge; my only thought on that is (without really any proof to back it up) visitors may trust a project more when they can see it runs with automated tests. It's not important though for now though, will remove. 

Let me try a few things to see if it is possible with composer to have it install the most appropriate PHPUnit version.

michaelbutler:
Looks like that did it! Now running fine on PHP 5.4 and PHP 5.5 too. This works because PHPUnit's API is backwards compatible but they changed it in 6.0 I believe.

PR #303, opened 2018-04-06
Add a few unit tests, clean up some minor issues

michaelbutler:
From looking at code coverage I've started adding a few tests to the unit suite to gradually cover things that weren't automatically tested before (e.g. the custom abbreviation expansion). 
I also started doing some minor tweaks such as changing == to === for string comparisons, adding `{ }` for if statements, and so on.


michelf:
This all looks fine to me. Will merge.

PR #307, opened 2018-07-05
Fix typo in Readme.php

bobbykjack:
autoload.php, not autoloader.php


michelf:
Oops! Thank you for the patch.

PR #309, opened 2018-08-08
Various little fixes

NicolasCARPi:
* Check if property is callable with is_callable everywhere needed
* Remove the & from the docblock declaration of variable passed by
reference
* Make default value of $footnotes_assembled an empty string instead of
null because it is concatenated afterwards with a string
* Use $mode instead of $this->mode (undeclared property anyway, and not
used anywhere else afaik)
* Correct the docblock types for some variables

Hello,

I'm using php-markdown in my project ([elabftw](https://github.com/elabftw/elabftw/)) and I decided to try and contribute back to the projects on which I depend.

So here is my humble contribution, a few fixes here and there :)

Thanks for this great library!

Cheers,
~Nico

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -85,25 +85,25 @@ class Markdown implements MarkdownInterface {
 
 	/**
 	 * Optional filter function for URLs
-	 * @var callable
+	 * @var callable|null
 	 */
 	public $url_filter_func = null;
 
 	/**
 	 * Optional header id="" generation callback function.
-	 * @var callable
+	 * @var callable|null
 	 */
 	public $header_id_func = null;
 
 	/**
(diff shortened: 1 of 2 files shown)


michelf:
Looks all good, except I disagree about `$footnotes_assembled`. For one thing it is a breaking change (someone could be checking for null explicitly). And secondly this variable should only have a value at all when `$omit_footnotes` is `false`, and I feel null represents better a "no value" value than an empty string.

NicolasCARPi:
Hello,

I have reverted the change related to `$footnotes_assembled`, even if I disagree because if it's a string, the default should imho be an empty string, it helps with static analysis to have non nullable variables with only one possible type. But the BC is a good argument. :)

Cheers,
~Nico

PR #311, opened 2018-09-11
Add syntax highlighting to code examples

NeonArray:


michaelbutler (approved):
looks good to me!

michaelbutler (changes requested):
Actually sorry, since you're adding the backticks, could you remove the indenting?

e.g.

```php
$foo = 'bar';
```

michelf:
Using the indented code block style (that can't specify which syntax to use) is done on purpose so the file is parsable by the plain `Markdown` parser. I'm not too sure I want to break that compatibility.

See previous discussion in pull request #254. Also, sorry for not chiming in earlier.

NeonArray:
Fair enough.

PR #314, opened 2018-12-11
Update Readme.php

skeptycal:
Added a file operation that gets the first line of the file so it can be used to set the page title.

### DIFF
--- a/Readme.php
+++ b/Readme.php
@@ -15,14 +15,27 @@ spl_autoload_register(function($class){
 use Michelf\Markdown;
 
 // Read file and pass content through the Markdown parser
-$text = file_get_contents('Readme.md');
+
+// Added $file variable for clarity
+$file = 'readme.md';
+$text = file_get_contents($file);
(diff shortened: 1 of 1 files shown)


michelf:
Since this file exists to showcase how to use the library, I'd rather not add code that distract from this. If getting the title of the document was a feature of the parser then it might make sense to showcase it there, but otherwise I'd avoid adding this to Readme.php.

Some comments about the code:

* The file name is "Readme.md", emphasis on the capital R.  Reading from "readme.md" will fail on case-sensitive file systems.

* Opening the file a second time using `SplFileObject` to extract the first line while the whole content of the file is already sitting in the `$text` variable is rather wasteful.

* I don't think this is a particularly good way to extract the title. Taking the first line will work with the current Readme.md document, but as a general rule the first line of a Markdown document could be blank leaving the title for a subsequent line, it could be an atx-style header starting with `#` (where `#` is _not_ part of the title), or it could have no title and the first line would be a paragraph, a list, or something else.

PR #316, opened 2019-07-26
curly braces is deprecated in PHP 7.4

remicollet:
`Deprecated:  Array and string offset access syntax with curly braces is deprecated in /usr/share/php/Michelf/Markdown.php on line xxx`

https://wiki.php.net/rfc/deprecate_curly_braces_array_access


michelf:
Thank you.

PR #322, opened 2019-12-28
Fixed failing travis builds

### LINKED ISSUE #324: Travis setup broken

GrahamCampbell:
Before: https://www.travis-ci.org/michelf/php-markdown/builds/619416215.

image

After: https://www.travis-ci.org/michelf/php-markdown/builds/630338990.

image

---

Closes #265. Closes #322. Closes #324.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,29 @@
 language: php
 
-sudo: false
+matrix:
+  include:
+    - php: hhvm-3.18
+      dist: trusty
+    - php: 5.3
+      dist: precise
+    - php: 5.4
+      dist: trusty
+    - php: 5.5
+      dist: trusty
+    - php: 5.6
+      dist: xenial
+    - php: 7.0
+      dist: xenial
+    - php: 7.1
+      dist: bionic
+    - php: 7.2
+      dist: bionic
+    - php: 7.3
+      dist: bionic
+    - php: 7.4
+      dist: bionic
 
-cache:
-  directories:
-    - $HOME/.composer/
-
-# Run tests against all these PHP versions
-# TODO: When it becomes possible in TravisCI, switch 7.4snapshot to plain 7.4
-php:
-  - 5.4
-  - 5.5
-  - 5.6
-  - 7.0
-  - 7.1
-  - 7.2
-  - 7.3
-  - 7.4snapshot
-  - hhvm
-
-before_script:
+install:
   - composer install --prefer-dist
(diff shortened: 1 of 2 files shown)


michelf:
Looks good to me. Although I don't see why the d62eb48 commit is relevant to the issue.

GrahamCampbell:
> Although I don't see why the d62eb48 commit is relevant to the issue.

Because the PHP 5.3 travis build fails without that change.

michelf:
Oh right. Sorry, I read the diff in the wrong direction, somehow I though you were replacing `array()` with `[]`. Makes sense now.

michelf:
I'd rather not drop PHP 5.3 it just for the sake of dropping it. But if something comes up that is not too practical with 5.3 I won't hesitate much to drop it.

cdp1337:
Fair enough.  I do really like many features that have come out in PHP 7.x, which break functionality with PHP 5, such as typecasting, strong return values, and the like.  For example 

```php
public static function defaultTransform($text) { ... }
```

with PHP 7 typecasting would be

```php
public static function defaultTransform(string $text) : string { ... }
```

I've found this extremely useful in some of my projects, and it may make sense to start mulling around the possibility of branching for a 7.x version in the next year or two to make use of these new features.  Certainly not a priority for a library like this, but just something to have in the back of your mind.

michaelbutler:
Sorry for outdated thread bump; but I have to agree with @cdp1337 -- by dropping 5.3 support, (or any 5.X version for that matter...) it does not mean we all of a sudden **break** anyone using this package on PHP 5.3. It simply means whatever code they have just continues to work fine for them. Assuming they are using composer correctly, that is, where the version would be locked at something like 1.X. This is because we increase the major version number (2.0), signifying a backwards-compatibility break.  This **does** mean that anything we change or add after the point of **dropping** support cannot be utilized by ancient PHP versions. This does not seem like a problem at all, as those ancient PHP installations are extremely unlikely to be continuing to receive composer updates anyway.

I think it's fully reasonable to say, if using < PHP 7.0, use php-markdown `1.*`, if on PHP 7.0+, use php-markdown `2.*`. 

That's my two cents :smile:

michelf:
@michaelbutler Perhaps if the project was more active it'd be different, but the maintenance burden for supporting 5.x is negligible at the moment. If you are planning to contribute some changes requiring a bump in the requirements, please open it as new issue.

PR #328, opened 2020-04-09
Remove blank title attributes from anchor tags

Verest:
Inlined anchor tags with no title attribute are being given a blank title in the case extra attributes are present.

### DIFF
--- a/Michelf/MarkdownExtra.php
+++ b/Michelf/MarkdownExtra.php
@@ -932,6 +932,7 @@ class MarkdownExtra extends \Michelf\Markdown {
 	protected function _doAnchors_inline_callback($matches) {
 		$link_text		=  $this->runSpanGamut($matches[2]);
 		$url			=  $matches[3] === '' ? $matches[4] : $matches[3];
+		$title_quote		=& $matches[6];
 		$title			=& $matches[7];
 		$attr  = $this->doExtraAttributes("a", $dummy =& $matches[8]);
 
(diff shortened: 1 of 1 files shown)


michelf:
I'm not too sure what this is supposed to fix. Are you saying you're observing that this:

    [some link](http://example.com)

produces a `title=""` attribute in the output? This is not what I observe on the [dingus](https://michelf.ca/projects/php-markdown/dingus/). The `title=""` attribute is added only if you explicitly write an empty title, such as in:

    [some link](http://example.com "")

Verest:
It only happens when the `preg_replace_callback` regex matches an additional attribute (the 8th captured group). For example...
```
[some link](http://example.com){target=_blank}
```
will be converted to...
```
<p><a href="http://example.com" title="" target="_blank">some link</a></p>
```
The goal is to remove the `title=""` portion.

michelf:
I see. That's a good fix. It's only a bit unfortunate that it removes the empty title if it was explicitly requested. I think you should check `$matches[6]` instead, which contains the quote character that surrounds the title: if it was left empty then no title was matched (not even an empty title).

Verest:
OK, the check is based off of the quote's existence now. I added it for images as well as it seems they had the same issue. 
.
I have the test `testPhpMarkdownMdTestWithMarkdownExtra` failing with dataset #8, but it also fails on the `lib` branch on my local machine so it doesn't seem related.

michelf:
Looking good. Thank you.

PR #329, opened 2020-04-22
Treat <details> and <summary> as block level.

th-h:
`<details>` and `<summary>` elements should never be wrapped in `<p></p>`.

Add test to `test\resources\php-markdown.mdtest\Inline HTML (Simple)` accordingly.


michelf:
Looks good. Thank you.

unsektor:
Hello, thank you for contributing, these changes prevents wrapping `<details>` element with `<p>` element, indeed.
Unfortunately, it looks like these changes are also causes a bug:
makdown content inside `<details>` and `<summary>` elements is not rendered properly anymore.
See #404 for details.

PR #352, opened 2021-04-07
Exclude tests from release archives

nikic:
In particular, this means that people using this package via composer do not download the test files, only the implementation.

If desired, I can also add other files not needed for production (readme, CI config etc).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/test export-ignore


michelf:
Merged #374 instead, which should exclude files from composer archives without excluding them from git archives (like when downloading the archive from GitHub.)

PR #355, opened 2021-07-08
Auto-generate preconfigured class attributes

fleonce:
From my understanding it is currently impossible to let the Markdown library generate class attributes in the html output. While it would be possible to include those after the text has been generated via regex replace, it is much easier to configure the Markdown object beforehand, e.g. if you want all images in the output to have the "control-img" class.

I dont know if this is rather a thing I would only need for my specific use-case or if there is more interest in this, I created a pull-request for it.

In my use-case, I generate forum article html content from Markdown source files, inside of the general php generated html content. This exposes me to the following problem:

If I include images in the Markdown, there is currently no possibility to control their width and height, although there is a pull request open for this [exact issue](https://github.com/michelf/php-markdown/pull/351). I believe allowing class attributes to be added to generated html tags to use them in css styling purposes would make things quite a bit easier.

I currently did only include attribute generation for html tags `a` and `img`, however if accepted I can of course add support for the other tags generated as well.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -125,6 +125,8 @@ class Markdown implements MarkdownInterface {
 	 * @var bool
 	 */
 	public $enhanced_ordered_list = false;
+	
+	public $class_attributes = null;
 
 	/**
 	 * Parser implementation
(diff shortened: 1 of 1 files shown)


michelf:
I don't understand why you'd want to put the same class names on every image. If all the images need the same style applied to them, isn't it already easy to target them with a CSS selector targeting an ancestor with a specific class? Like `div.post img { width: 100% }` Am I missing something?

On the implementation side, I think this won't work with [special attributes](https://michelf.ca/projects/php-markdown/extra/#spe-attr) in the `MarkdownExtra` parser.

PR #356, opened 2021-08-15
Create sample1.md

rheda-cx:


michelf:
I'm not too sure what's the point of adding this test case. Does it test for a an existing bug?

And if it's a test case, where is the corresponding file for the expected result?

PR #358, opened 2021-11-06
Bump to php7.4

tacman:
Change the minimum requirements, and change the travis tests.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,7 +1,7 @@
 build:
   environment:
     php:
-      version: '7.0.20'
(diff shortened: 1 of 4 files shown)


michelf:
Can you also update [this line in .scrutinizer.yml](https://github.com/michelf/php-markdown/blob/lib/.scrutinizer.yml#L4) so the analysis pass is not [broken](https://scrutinizer-ci.com/g/michelf/php-markdown/inspections/e56dba26-33ff-4477-86ee-afb6ffcf2d39)?

Issue #333, opened 2020-10-27
Recursion limits hit more often in PHP 7.3+


michaelbutler:
With the major release of PHP 7.3 there were some [big changes under the hood](https://hackernoon.com/deprecations-and-changes-for-php-7-3-avoid-errors-7a5c4dbeaa8b) with the regular expression engine, going from PCRE to PCRE2. While it is not explicitly called out in the changes, I noticed that this new PCRE2 engine is a lot more susceptible to the `PREG_RECURSION_LIMIT_ERROR` error when using `MarkdownExtra`. Let's dig in to the details.

`pcre.recursion_limit` is the [INI setting](https://www.php.net/manual/en/pcre.configuration.php) that controls:

> (Default 100000) PCRE's recursion limit. Please note that if you set this value to a high number you may consume all the available process stack and eventually crash PHP (due to reaching the stack size limit imposed by the Operating System).

In PHP 7.2 and below, we had actually decreased the 100,000 default to 10,000 for performance reasons (to stop out-of-control user content from slowing down PHP).  But in PHP 7.3, having this 10,000 number now fails on pieces of content that did not fail before. This led me to think that PHP 7.3 is more susceptible to this.

## What happens when the recursion limit is hit? 

Simple, the `preg_split` call returns `false`, which [this library doesn't expect](https://github.com/michelf/php-markdown/blob/lib/Michelf/MarkdownExtra.php#L514), so the code fails at a later line due to unexpected data type (`false` is not Countable, also a fairly recent & strict PHP error).

## Example Content

But what specifically is in the content that causes us to hit this limit? It turns out that it has to do with dangling `<` in the markdown -- for example,  `<3` for a heart symbol, or doing a math comparison like `4 < 8` or really any other `<` that is not part of an HTML tag. That alone isn't a problem, you also need to have "a lot" of content after the `<`. I tested truncating the markdown to various lengths and it wouldn't fail until it was sufficiently long enough. For the INI setting of 10,000 this failed at a document length of around 5k chars where the `<3` was near the top. In fact I'll attach the file here [testpost.txt](https://github.com/michelf/php-markdown/files/5446581/testpost.txt). to test, run with:

```
echo "<?php  \Michelf\MarkdownExtra::defaultTransform(file_get_contents('testpost.txt'));" > testing.php;
php -dauto_prepend_file=vendor/autoload.php -dpcre.recursion_limit=10000 testing.php
```

[...]

PR #359, opened 2021-11-07
Typed Properties

tacman:
Change all the annotations to typed properties, e.g.

```php
	/**
	 * Optional title attribute for footnote links.
	 * @var string
	 */
	public $fn_link_title = "";
```

to 

```php
	/**
	 * Optional title attribute for footnote links.
	 */
	public string $fn_link_title = "";
```

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -31,7 +31,7 @@ class Markdown implements MarkdownInterface {
 	 */
 	public static function defaultTransform($text) {
 		// Take parser class on which this function was called.
-		$parser_class = \get_called_class();
+		$parser_class = static::class;
 
 		// Try to take parser from the static parser list
 		static $parser_list;
@@ -49,25 +49,21 @@ class Markdown implements MarkdownInterface {
 	/**
 	 * Configuration variables
 	 */
-
 	/**
 	 * Change to ">" for HTML output.
-	 * @var string
 	 */
-	public $empty_element_suffix = " />";
+	public string $empty_element_suffix = " />";
 
 	/**
 	 * The width of indentation of the output markup
-	 * @var int
 	 */
-	public $tab_width = 4;
+	public int $tab_width = 4;
 
 	/**
 	 * Change to `true` to disallow markup or entities.
-	 * @var boolean
 	 */
-	public $no_markup   = false;
-	public $no_entities = false;
+	public bool $no_markup   = false;
+	public bool $no_entities = false;
 
 
 	/**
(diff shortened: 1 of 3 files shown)


tacman:
Should we tag this as a version 2?  And maybe keep it in dev?  The tests pass, but I'm wondering if it'll be a problem for users who are upgrading.

michaelbutler:
definitely will want to tag this as 2.0, and update the Readme.md to say that the latest version requires PHP 7.4 or higher, while 1.9 and below support back to 5.3.

michelf:
That version 2 bump seems right, but after testing things now I'm a bit confused. I thought those type hints would cause errors for things like:
```
use Michelf\MarkdownExtra;
$mde = new MarkdownExtra;
$mde->hard_wrap = "zzz"; // should be a boolean!
echo $mde->transform(131); // should be a string!
```
But, unexpectedly for me, this runs fine — PHP 8.0.12 (cli) — so now I wonder where's the issue. I must be missing something.

michaelbutler:
@michelf I'm not too certain myself, but we might need to add this to the top of the file(s) to get those to error:

`<?php declare(strict_types=1);`

Which sounds scary, but actually this only affects logic that happens within the file, and calls to external classes/objects/methods. In other words, enabling this WON'T cause problems if downstream usages DON'T have strict types enabled. But it does mean that we would need to be extra sure we're not doing any invalid usages within the file, and we probably would need to test it a bit more first.

https://stackoverflow.com/questions/37111470/enabling-strict-types-globally-in-php-7/37112026

michelf:
Ok, now I see. Thank you for the link.

I'm mostly interested in the effects for users of the library at the moment so we can answer the question of whether it should be bumped to v.2. Looks like only a user declaring `strict_types=1` in the file that calls or configures the Markdown parser can be affected, and only if the user is also relying on type coercion (like doing `$mde->hard_wrap = 1`). At first glance that sounds extremely unlikely in the same file, but rewriting it as `$mde->hard_wrap = read_config_from_somewhere()` makes it sound more likely. I'll give it some more thoughts.

Note that the versioning policy [excludes protected members](https://github.com/michelf/php-markdown#public-api-and-versioning-policy) from semantic versioning, so I'm only considering the effects of type hints for the public members.

michaelbutler:
If we don't bump to 2.0 we risk syntax errors for those who used `composer require 1.*` to require the library. They might accidentally upgrade and if they're on PHP 7.3 or lower it will be an unrecoverable error.

tacman:
Yes -- like KnpMarkdownBundle, which is a Symfony wrapper for php-markdown.

michelf:
Would composer actually do the upgrade without checking the PHP 7.4 requirement is satisfied?

In any case, I think this will be a v.2. It'll make it clear that the PHP requirement has changed. It might also allow some changes to the public API, if necessary.

michaelbutler:
Good point, it might be smart enough to just skip the upgrade due to the PHP requirement. Or it could error, I don't know. 

I could also address https://github.com/michelf/php-markdown/issues/333 for version 2 as well (adding exceptions, or at least just fixing the case where preg_* function could return `false`)

PR #368, opened 2022-02-18
Fix PHPStan issues and build on PHP 8.2

sanmai:
Looks like PHPStan is kinda lost here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                php: 
+                php:
                   - '7.4'
                   - '8.0'
                   - '8.1'
@@ -26,4 +26,5 @@ jobs:
             -   run: composer install --no-progress
 #            -   run: composer codestyle
             -   run: composer phpstan
+                if:  matrix.php == '8.1'
             -   run: composer tests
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -24,11 +24,10 @@
  		"psr-4": { "Michelf\\": "Michelf/" }
 	},
 	"require-dev": {
-		"phpspec/prophecy": "^1.6",
```
sanmai, line 27: There's no need for this dependency. See:
https://github.com/sebastianbergmann/phpunit/issues/5033#issuecomment-1221992857

michelf:
I see you're hitting some problems with the PHPStan and the tests. I think it's confused by the implicit treatment of DOMNode as DOMText (and perhaps other similar things). I tried to fix it in #377, but ended up fixing the wrong things and abandoned because I had not time left for this.

Thank you for your work.

.github/workflows/ci.yml
```diff
@@ -26,4 +26,5 @@ jobs:
             -   run: composer install --no-progress
 #            -   run: composer codestyle
             -   run: composer phpstan
+                if:  matrix.php == '8.1'
```
sanmai, line 29: If we run PHPStan on every version, we'll use different versions of PHPStan detecting different sets of issues (due to bugs in older versions or whatnot).

sanmai:
@michelf Thanks for considering this PR! Please let me know if you need anything else or something isn't clear. Hopefully you can squash-merge this mess of commits.

sanmai:
Build is passing :tada:

PR #382, opened 2022-09-22
Remove type attributes

sanmai:
As per https://github.com/michelf/php-markdown/pull/379#issuecomment-1254979771

To be clear, I would seriously _love_ to see a next major version with type annotations. But if there's a choice between getting a minor version bump and no release, I'd prefer to have the release pipeline going. If this makes any sense.

Please feel free to close this PR if you'd rather stick with typed properties.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -18,7 +18,7 @@ class Markdown implements MarkdownInterface {
 	 * Define the package version
 	 * @var string
 	 */
-	const MARKDOWNLIB_VERSION = "2.0";
+	const MARKDOWNLIB_VERSION = "1.9.2";
 
 	/**
 	 * Simple function interface - Initialize the parser and return the result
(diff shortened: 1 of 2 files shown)


michelf:
I think per the [versioning policy](https://github.com/michelf/php-markdown/blob/lib/Readme.md#public-api-and-versioning-policy) we should keep the types attributes for protected properties. 

For the public ones, this is more nuanced, and I'm still trying to wrap my head around it. Strictly speaking this is a breaking change for users using `declare(strict_types=1)`. Even without this mode, it could in some case create errors where a user of the library would be setting a string property with an array value, or the reverse (which might or might not create an error down the line depending on the parsed input). In practice I doubt very much it'd affect anyone. It all seems very unlikely to me, but I don't really know. Any comment on that?

In any case I don't think the downsides of making it a major version (slowing down adoption, prompting people to read a changelog saying "we added type attributes" before deciding to allow a 2.0.0 version) is worth it.

sanmai:
Sounds great to me! Following this with #383

PR #384, opened 2022-09-26
v1.10.0

michelf:
Not entirely sure yet if this should be 2.0.0 or 1.10.0, but I'm leaning for 1.10.0 because it seems very unlikely type annotations would break any code and the new requirement for PHP 7.4 should be handled fine by composer.


michelf:
Abandoned, going to use 2.0.0 after additional type declarations in `MarkdownInterface`. See #385

Issue #326, opened 2020-01-17
Links are not automatically created


qwwwest:
GitHub automatically creates links from standard URLs...
but php-markdown does not do it unless the link is placed between '<' and '>'

michelf:
[The original Markdown syntax](https://daringfireball.net/projects/markdown/syntax#autolink) does not automatically link text that looks like a link. This is an extension of Github-flavored-Markdown.

recolic:
Could we add a switch, to allow user to enable this feature/extension?

michelf:
Adding it as a configurable option seems like a good idea. Someone wants to make a pull request?

appel:
I just created a quick pull request that adds a switch to automatically create links. Simple, but seems to do the job. Hope this is useful!

PR #386, opened 2022-10-25
Added a switch to enable autolinking of links.

appel:
As discussed in https://github.com/michelf/php-markdown/issues/326. It's not pretty but it seems to work.

### DIFF
--- a/Michelf/Markdown.php
+++ b/Michelf/Markdown.php
@@ -65,6 +65,11 @@ class Markdown implements MarkdownInterface {
 	public bool $no_markup   = false;
 	public bool $no_entities = false;
 
+	/**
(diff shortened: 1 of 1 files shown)


edent:
I like this, but I think it breaks if there's a link inside another link.

For example:

```
Markdown [Visit https://example.com](https://example.com)
HTML <a href="https://example.com">Visit https://example.com</a>
```

Is rendered as:

```html
<a href="https://example.com">Visit <a href="https://example.com">https://example.com</a></a>
```

appel:
Hey edent, thanks for your comment. Honestly, I can't remember if I tested this at the time, but I agree that that would be a bit of an issue. :) I don't think my PR really went anywhere and I have since moved on to a different MD implementation, so I won't be able to look into this, unfortunately.

Issue #387, opened 2022-11-15
Better Footnote title attributes?


edent:
Footnote links' titles can only be set once by the library. This change would make the attribute display the content of the footnote.

### How

At the moment, if a user hovers over a footnote, the `title` attribute displays a tool tip.

For example, if 
https://github.com/michelf/php-markdown/blob/eb176f173fbac58a045aff78e55f833264b34e71/Michelf/MarkdownExtra.php#L28
is set to "Read the footnote." then *all* footnotes have this pop-up:

<img src="https://user-images.githubusercontent.com/837136/201670341-abb02a98-0082-493e-bf1d-50d69b09e765.png" alt="Screenshot of a pop up which says 'read the footnote'" width="320" />

<hr>

I think it would be nice if it showed users the *full* text of the footnote. For example:

<img alt="The pop up now displays the full note" src="https://user-images.githubusercontent.com/837136/201670338-99a41e45-2168-419c-91ef-183ca2d1a867.png" width="320" />
<hr>

This is controlled by :

https://github.com/michelf/php-markdown/blob/eb176f173fbac58a045aff78e55f833264b34e71/Michelf/MarkdownExtra.php#L1764

The change is relatively straightforward:

```php
if ($this->fn_link_title != "") {
	$title = trim( strip_tags( $this->footnotes[$node_id] ) );
	$title = $this->encodeAttribute( $title );
	$attr .= " title=\"$title\"";
}
```

That takes the text of the footnote, sanitises it, and adds it as the title element.

Would you be interested in a PR for this?

(I have also raised this with WordPress's Jetpack which is running an older version of your library - https://github.com/Automattic/jetpack/issues/27387)

michelf:
How do you propose to handle footnotes like this one[^1].

[^1]: Footnotes are often one line of text, but this line of text can contain formatting such as `code`, _emphasis_, [links](https://michelf.ca/php-markdown/), and images ![such as this one](https://user-images.githubusercontent.com/298045/201915639-7f0f0ff5-8ce6-4c6e-a54f-4fbf3fead7b2.png). Actually, it wouldn't surprise me to find that links are pretty common since footnotes are often used for references.

    They can also go quite long with paragraphs and other block-level elements such as:
    
    | tables | tables |
    | ------ | ------ |
    | rows   | rows   |
    
    > blockquotes

        code blocks

    - lists
    - and more lists

    So we need a strategy to deal with this. I don't think `strip_tags` is good enough.

[...]

PR #390, opened 2023-03-26
Titles for footnotes

edent:
Re #387 - I've created a function which adds truncated title text to footnote links, and I've added some tests to show what it does.

Very happy for any comments or corrections.


test/resources/php-markdown-extra.mdtest/Footnotes.xhtml
```diff
@@ -82,20 +100,76 @@ three lines, with some span-level markup like
 </li>
 
 <li id="fn:reference" role="doc-endnote">
-<p>This footnote has a footnote of its own.<sup id="fnref:nested"><a href="#fn:nested" class="footnote-ref" role="doc-noteref">12</a></sup>&#160;<a href="#fnref:reference" class="footnote-backref" role="doc-backlink">&#8617;&#xFE0E;</a></p>
+<p>This footnote has a footnote of its own.<sup id="fnref:nested"><a href="#fn:nested" class="footnote-ref" title="This footnote should appear even though it is referenced from another footnote. But Ffn:reference: should be literal since the footnote with that name has already been used." role="doc-noteref">20</a></sup>&#160;<a href="#fnref:reference" class="footnote-backref" role="doc-backlink">&#8617;&#xFE0E;</a></p>
```
michelf, line 103: I think "But Ffn:reference: should" isn't the intended result here.

test/resources/php-markdown-extra.mdtest/Footnotes.xhtml
```diff
@@ -1,34 +1,52 @@
...
+<li>List item one.<sup id="fnref:second"><a href="#fn:second" class="footnote-ref" title="This is the second note." role="doc-noteref">2</a></sup></li>
+<li>List item two.<sup id="fnref:third"><a href="#fn:third" class="footnote-ref" title="This is the third note, defined out of order." role="doc-noteref">3</a></sup></li>
 </ul>
 
-<h1>Header<sup id="fnref:fourth"><a href="#fn:fourth" class="footnote-ref" role="doc-noteref">4</a></sup></h1>
+<h1>Header<sup id="fnref:fourth"><a href="#fn:fourth" class="footnote-ref" title="This is the fourth note." role="doc-noteref">4</a></sup></h1>
 
-<p>Some paragraph with a footnote<sup id="fnref:1"><a href="#fn:1" class="footnote-ref" role="doc-noteref">5</a></sup>, and another<sup id="fnref:2"><a href="#fn:2" class="footnote-ref" role="doc-noteref">6</a></sup>.</p>
+<p>Some paragraph with a footnote<sup id="fnref:1"><a href="#fn:1" class="footnote-ref" title="Content for fifth footnote." role="doc-noteref">5</a></sup>, and another<sup id="fnref:2"><a href="#fn:2" class="footnote-ref" title="Content for sixth footnote spanning on three lines, with some span-level markup like emphasis, a link." role="doc-noteref">6</a></sup>.</p>
```
michelf, line 10: Shouldn't this footnote be on three lines?

test/resources/php-markdown-extra.mdtest/Footnotes.xhtml
```diff
@@ -1,34 +1,52 @@
...
 
 <p>This paragraph should not have a footnote marker since 
 the footnote is undefined.[^3]</p>
 
-<p>This paragraph has a second footnote marker to footnote number one.<sup id="fnref2:1"><a href="#fn:1" class="footnote-ref" role="doc-noteref">5</a></sup></p>
+<p>This paragraph has a second footnote marker to footnote number one.<sup id="fnref2:1"><a href="#fn:1" class="footnote-ref" title="Content for fifth footnote." role="doc-noteref">5</a></sup></p>
 
 <p>This paragraph links to a footnote with plenty of 
-block-level content.<sup id="fnref:block"><a href="#fn:block" class="footnote-ref" role="doc-noteref">8</a></sup></p>
+block-level content.<sup id="fnref:block"><a href="#fn:block" class="footnote-ref" title="Paragraph. * List item > Blockquote Code block" role="doc-noteref">8</a></sup></p>
```
michelf, line 20: Does it really make sense to flatten block-level markup like this? At a minimum it should take three lines (using newline characters). But personally, my inclination would be to truncate after the first paragraph instead of doing a textual approximation of lists, blockquotes or other block-level markup.

test/resources/php-markdown-extra.mdtest/Footnotes.xhtml
```diff
@@ -1,34 +1,52 @@
...
 
 <p>This paragraph host the footnote reference within a 
-footnote test<sup id="fnref:reference"><a href="#fn:reference" class="footnote-ref" role="doc-noteref">9</a></sup>.</p>
+footnote test<sup id="fnref:reference"><a href="#fn:reference" class="footnote-ref" title="This footnote has a footnote of its own.Ffn:nested:" role="doc-noteref">9</a></sup>.</p>
```
michelf, line 23: I don't think "Ffn:nested:" is what we want here. (What do we want for nested footnotes is a good question.)

test/resources/php-markdown-extra.mdtest/Footnotes.xhtml
```diff
@@ -1,34 +1,52 @@
...
+<img src="images/MGR-1800-travel.jpeg" alt="1800 Travel" title="Travel Speeds in 1800" />
+<img src="images/MGR-1830-travel.jpeg" alt="1830 Travel" title="Travel Speeds in 1830" /></p>
+
+<p>Alt text in images is exposed in the title<sup id="fnref:imgalt"><a href="#fn:imgalt" class="footnote-ref" title="This is snuggle. Photo of a cute kitten." role="doc-noteref">12</a></sup>.</p>
+
+<p>Very long footnotes are truncated at word boundaries<sup id="fnref:long"><a href="#fn:long" class="footnote-ref" title="Sensors indicate no shuttle or other ships in this sector. According to coordinates, we have travelled 7,000 light years and are located near the system J-25. Tractor beam released, sir. Force field…" role="doc-noteref">13</a></sup>.</p>
+
+<p>Although some titles may be split-mid emoji if they use ZWJ<sup id="fnref:zwj"><a href="#fn:zwj" class="footnote-ref" title="👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍🦲👨‍…" role="doc-noteref">14</a></sup>.</p>
```
michelf, line 39: Splitting characters is unfortunate. And I suppose the same thing could happen to any unicode character composed of multiple code points since the string is sliced on a code point boundary. So this is most likely not only a problem for emoji.

test/resources/php-markdown-extra.mdtest/Footnotes.xhtml
```diff
@@ -1,34 +1,52 @@
...
+
+<p>Fully multilingual<sup id="fnref:中国文"><a href="#fn:中国文" class="footnote-ref" title="两位皇子，你们各拥党羽，雄心勃勃地争取国柄和皇座，我们现在代表民众告诉你们：罗马人民已经众口一辞，公举素有忠诚之名的安德洛尼克斯作为统治罗马的君王，因为他曾经为罗马立下许多丰功伟绩，在今日的邦城之内，没有一个比他更高贵的男子，更英勇的战士。他这次奉着元老院的召唤，从征讨野蛮的哥特人的辛苦的战役中回国；凭着他们父子使敌人破胆的声威，已经镇伏了一个强悍善战的民族。自从他为了罗马的光荣开始出征、用武力膺…" role="doc-noteref">15</a></sup>.</p>
+
+<p>Also supports elements like <code>&lt;ruby&gt;&lt;rp&gt;&lt;rt&gt;</code><sup id="fnref:ruby"><a href="#fn:ruby" class="footnote-ref" title="漢 (kan) 字 (ji)" role="doc-noteref">16</a></sup></p>
+
+<p>Also Emoji<sup id="fnref:🥰"><a href="#fn:🥰" class="footnote-ref" title="🥰" role="doc-noteref">17</a></sup> compatible.</p>
+
+<p>All the text present in a table should be in the title tooltip<sup id="fnref:table"><a href="#fn:table" class="footnote-ref" title="The table header The table body with two columns Footer A demo table" role="doc-noteref">18</a></sup>.</p>
```
michelf, line 47: A better approximation would be to show the table with `|` separators between columns and a newline after each row. Personally, my choice would be to not offer a title attribute here, forcing people to click the link to see a proper table.

Michelf/MarkdownExtra.php
```diff
@@ -1760,8 +1842,10 @@ protected function _appendFootnotes_callback($matches) {
 				$class = $this->encodeAttribute($class);
 				$attr .= " class=\"$class\"";
 			}
-			if ($this->fn_link_title !== "") {
-				$title = $this->fn_link_title;
+			if ($this->fn_link_title == "") {
```
michelf, line 1845: This logic doesn't make much sense. If a title was specified by `fn_link_title` it shouldn't be ignored. There should be an `else` clause.

Michelf/MarkdownExtra.php
```diff
@@ -1760,8 +1842,10 @@ protected function _appendFootnotes_callback($matches) {
...
+				//	Decode any markdown in the footnote
+				$title = $this->get_footnote_title( $this->formParagraphs( $this->footnotes[$node_id] ) );
+				//	Format it to be suitable for a title tool-tip
 				$title = $this->encodeAttribute($title);
```
michelf, line 1849: I'm not sure `encodeAttribute` will handle correctly your generated footnote title here when the `no_entities` configuration variable is set to `true`. I haven't checked, but I think it could double encode `&` and `<` into `&amp;amp;` and `&lt;lt;`.

Michelf/MarkdownExtra.php
```diff
@@ -1760,8 +1842,10 @@ protected function _appendFootnotes_callback($matches) {
 				$class = $this->encodeAttribute($class);
 				$attr .= " class=\"$class\"";
 			}
-			if ($this->fn_link_title !== "") {
-				$title = $this->fn_link_title;
+			if ($this->fn_link_title == "") {
+				//	Decode any markdown in the footnote
+				$title = $this->get_footnote_title( $this->formParagraphs( $this->footnotes[$node_id] ) );
```
michelf, line 1847: It's unfortunate we're calling `formParagraph twice for every footnote now. Once for the footnote reference, and once for the actual footnote at the end of the document. More than twice if there is more than one references to the same footnote. I think this requires more strategy.

michelf (changes requested):
I'm a little worried about the performance of this approach for `get_footnote_title`. Maybe it's fast and I'm worrying too much, but I've always tried to keep the parser extremely fast so the library can be used without caching in many situations. To me, running the HTML parser to create a DOM, searching the DOM to replace various things, then rewriting that DOM to HTML, stripping the HTML tags, splitting on every space and reassembling those words, and calling `mb_strlen` in a loop, is not very efficient. So I'm wary and I guess I'll want to see some before/after performance benchmarks, especially with long footnotes, before accepting it to be sure this is not a regression in speed.

I also see tests are failing in CI. Are you sure the output is well-formed XHTML?

PR #393, opened 2023-04-27
Add id and class attr for table

iwill:
Add id and class attr for table in Markdown:

```md
| ID | Name   | State |
| -: | :----- | :---: |
|  1 | Task A | DONE  |
|  2 | Task B | TODO  |
{ #id .class }
```

HTML:

```html
<table id="id" class="class">
   ...
</table>
```

### DIFF
--- a/Michelf/MarkdownExtra.php
+++ b/Michelf/MarkdownExtra.php
@@ -1154,6 +1154,7 @@ class MarkdownExtra extends \Michelf\Markdown {
 						[|] .* \n			# Row content.
 					)*
 				)
+				('.$this->id_class_attr_catch_re.')?	 # $4 = id/class attributes
 				(?=\n|\Z)					# Stop at final double newline.
 			}xm',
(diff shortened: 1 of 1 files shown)


iwill:
@caiyili Can you fix the errors? 😅

michelf:
I see in the logs:
```
Undefined offset: 4

/home/runner/work/php-markdown/php-markdown/Michelf/MarkdownExtra.php:1199
```
It seems the error comes from a notice about `$matches[4]` when there is a table with no extra attribute block. I suggest you use the null coalescing operator (`$matches[4] ?? null`) to handle more cleanly cases where the regular expression finds no extra attribute block and there is no `$matches[4]`.

iwill:
@michelf Hi, I fixed the unit test error, would you like to merge this PR?

michelf:
@iwill Looks good. Sorry about it taking so long. I think GitHub never notified me about you pushing more changes.

PR #403, opened 2024-09-19
Fix php deprecated notice improving handling of null parsed attr values

detain:
Fix php deprecated notice improving handling of null parsed attr values


michelf:
Merged #365 instead of this one.

